PR #7, opened 2015-03-22
git submodule update is required after git submodule init

skitt:
At least on my system, with git 2.1.4, git submodule init doesn't clone the submodules; git submodule update is required to get the third-party source code.


dkopecek:
Hello! You're right, I totally forgot that. Thanks!

PR #53, opened 2016-02-12
Match icon name

genodeftest:
Currently usbguard doesn't have an icon because `Icon=` specifies a wrong name.


dkopecek:
Thanks! Merged.

PR #70, opened 2016-05-26
install service file without executable bit set

murinicanor:
thanks for usbguard!

the systemd service file should be installed without the executable bit set


dkopecek:
Hello and many thanks for your contributions!

PR #72, opened 2016-05-26
fix typos in the usbrules manpage

murinicanor:
the usbguard-rules manpage said usbguard-daemon instead of usbguard-rules in some places and one formated should be formatted ;)


dkopecek:
Thanks!

PR #77, opened 2016-05-29
Manpage generation from source

murinicanor:
hi,

it would be great if you would distribute the sources for the manpages in the releases. i could then (re)create the *.roff files from source.
i've added a small snipplet to create manpages with _ronn_, which is both in debian and in fedora (in contrast to _ronn-nodejs_)

cheers,
muri


dkopecek:
Ok. I agree with the idea. I somehow missed that the utility (or rather a similar one) is packaged in Fedora.

I have a couple of questions: 

1. I'm not entirely sure whether your changes are enough. Could you elaborate more on what is the desired end state here? I would expect an additional commit that removes the generated roff files from the repo. Or did you intent to leave them there and remove the files each time before you initiate a build of the package?

2. If you look at https://github.com/dkopecek/usbguard/blob/master/doc/README.md, the command for generating the roff files contain also a specification of the numeric section for the manual page. Did you forget to add it or is that handled in a different way with the ronn utility?

murinicanor:
> I'm not entirely sure whether your changes are enough. Could you elaborate more on what is the desired end state here? I would expect an additional commit that removes the generated roff files from the repo. Or did you intent to leave them there and remove the files each time before you initiate a build of the package?

yeah, i wasn't sure how you want to handle that- i thought you maybe want to still have the generated *.roff files to be part of the release.
i've added a commit that removes the *.roff files.

> If you look at https://github.com/dkopecek/usbguard/blob/master/doc/README.md, the command for generating the roff files contain also a specification of the numeric section for the manual page. Did you forget to add it or is that handled in a different way with the ronn utility?

apparently the ronn utility does this by looking at the filename, i've updated the commit accordingly. this lead to the file extension _.roff_ being in the way of automake and resulting in filenams like _usbguard.1.1_ after installing ("If the file already has a valid suffix, then it is installed as-is; otherwise the file suffix is changed to match the section."), so i've removed the _.roff_ extension in Makefile.am

dkopecek:
It seems that the changes break CI tests:
```
make[2]: *** No rule to make target `doc/usbguard.1.roff', needed by `all-am'.  Stop.
make[2]: Leaving directory `/home/travis/build/dkopecek/usbguard/build'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/home/travis/build/dkopecek/usbguard/build'
make: *** [all] Error 2
```

dkopecek:
btw. would you be able to squash the commits into one? So that the history is not polluted with commits that fix some commits before.

murinicanor:
> btw. would you be able to squash the commits into one?

yes, done!

> It seems that the changes break CI tests:

thats fixed now- one problem still: when i add ruby-ronn to the packages installed in travis.yml, it seems there is a problem with runnin ronn in the ci-test-environment, so i've disabled that for now (which means it doesn't generate the manpages). don't know how to deal with that...

dkopecek:
I don't think we can just leave it disabled. I propose the following solution: 

* Leave the generated man-pages in the distribution for cases where `ronn` isn't available on the build host.
* The source path of the man-pages in the `Makefile.am` will point to a *builddir* (read-write), not a *srcdir* (read-only during distcheck).
* If `ronn` isn't available, then the generating step will call just `cp` on the shipped files.
* If `ronn` is available, then the man-pages will be generated from the source files in *srcdir* into the *builddir*.

What do you think?

murinicanor:
> What do you think?

yes, thats a very sensible approach. i'll look into it in the next view days!

murinicanor:
sorry, but i'm not proficient enought with automake to implement this. i've looked into vpath builds but i don't know how to handle that ;(

dkopecek:
@murinicanor ok, no problem, I'll look into that

dkopecek:
This will be fixed via #83.

PR #83, opened 2016-06-05
Generate man-pages from source during the build phase, if possible.

dkopecek:
Otherwise use the pre-gererated versions distributed in the
tarball/repository when installing them during make install.


dkopecek:
Hmm, looks like the ruby-ronn package on Ubuntu trusty doesn't work:
```
...
Setting up ruby-fast-xs (0.8.0-3build1) ...
Setting up ruby-hpricot (0.8.6-5ubuntu1) ...
Setting up ruby-rack (1.5.2-1) ...
Setting up ruby-tilt (2.0.0+really1.4.1-1) ...
Setting up ruby-rack-protection (1.5.1-1) ...
Setting up ruby-sinatra (1.4.3-2) ...
Setting up ruby-mustache (0.99.4-4) ...
Setting up ruby-rdiscount (1.6.8-3build1) ...
Setting up ruby-ronn (0.7.3-2) ...
...
ronn --roff --pipe ../doc/usbguard.1.ronn > ./doc/usbguard.1.roff
/usr/local/rvm/rubies/ruby-2.2.3/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- rdiscount (LoadError)
	from /usr/local/rvm/rubies/ruby-2.2.3/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/bin/ronn:46:in `<main>'
make[1]: *** [doc/usbguard.1.roff] Error 1
```
@murinicanor I think the PR is ok but I won't merge it until I figure what is the problem here and fix it. Feel free to use the patch for the package.

dkopecek:
It looks like it might be possible to use pandoc to generate the man pages from markdown. That's probably the most portable option.

murinicanor:
thanks! that works great, and pandoc is definitely the most common used solution

PR #85, opened 2016-06-23
Reimplement descriptor parsing

dkopecek:
We used a quite naive way to parse USB descriptors which contained
wrong assumptions and could handle only simple cases of descriptor
hierarchies. For more complicated ones, the parser would return
garbage results.

This commit introduces a better way of parsing the data which is
based on the fact that every descriptor contains a common header
consisting of two fields -- the length of the descriptor and the
type. The length value cannot be trusted, so there are sanity
checks inside the core parser and any detected deviation from the
normal/sane will result in an exception. Furthermore, the parsing
can now be easily extended to support other types of USB descriptors.

The usbguard CLI tool has been extended with a new subcommand that
can be used for parsing binary USB descriptor files and displaying
them in a human-readable/textual form.


rebecca-palmer:
Works for me.

To avoid future confusion/bugs, I suggest adding a comment that _num_interfaces (src/Library/DevicePrivate.hpp:71, and currently unused except for debug messages) is *not* necessarily the length of _interface_types, as interfaces with alternate settings have a separate descriptor for each.

dkopecek:
Thanks for the feedback! I've deleted the variable, it was useless.

PR #90, opened 2016-07-12
Update usbguard-applet-qt.desktop.in

murinicanor:
add keywords to desktop file

see also https://wiki.gnome.org/Initiatives/GnomeGoals/DesktopFileKeywords
and https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s05.html


dkopecek:
Thanks!

PR #99, opened 2016-07-31
configure.ac: add systemwide pegtl include path

murinicanor:
this adds the include path for pegtl if the systemwide pegtl is used


dkopecek:
Hello. Is this change really needed? PEGTL header files have a main-header-file + header files in sub-directory structure. Upstream doesn't say how to install them but I would say that you should end up with:

/usr/include/pegtl.hh
/usr/include/pegtl/ { sub-directory header files }

then you don't need the -I/usr/include/pegtl because you should use the pegtl/ sub-directory prefix anyway when including files that aren't included by the main pegtl.hh, for example:

https://github.com/ColinH/PEGTL/blob/master/examples/uri_trace.cc

murinicanor:
hm, you're totally right! i'll have to check why it doesn't find the pegtl files without that change in my setup.
closing ;)

PR #100, opened 2016-07-31
Makefile.am: remove CPPFLAGS from moc

murinicanor:
moc doesn't know `-Wdate-time` which is part of the CPPFLAGS on debian to achieve reproducable builds (https://gcc.gnu.org/gcc-4.9/changes.html#languages: "warns when the __DATE__, __TIME__ or __TIMESTAMP__ macros are used. Those macros might prevent bit-wise-identical reproducible compilations.")
i'm not sure if there are CPPFLAGS needed by moc. also not sure if this could also happen with AM_CPPFLAGS

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -468,7 +468,7 @@ usbguard_applet_qt_CPPFLAGS=\
 	-fPIC @qt5_CFLAGS@
 
 %.moc.cpp: %.h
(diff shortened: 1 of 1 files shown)


dkopecek:
Hello. You are right, CPPFLAGS shouldn't be there as well as AM_CPPFLAGS. Please update the PR to remove AM_CPPFLAGS too.

murinicanor:
> Please update the PR to remove AM_CPPFLAGS too.
done!

dkopecek:
Thanks!

PR #103, opened 2016-08-06
processDeviceInsertion: Reject only if an exception occurred, not always

rebecca-palmer:
Commit d15d5fac50f419021e6835499901e50d43119e46 moved the "reject device" (intended for devices that trigger an exception, e.g. malformed descriptor) in processDeviceInsertion out of the catch block; this causes *all* newly inserted devices (even allowed ones) to be rejected.


dkopecek:
Thanks!

PR #126, opened 2016-09-08
fix build on mips

murinicanor:
this came up during the automated build for debian on mips (see bugreport https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=836712)

Christian Seiler explained the problem and provided the patch (https://lists.debian.org/debian-mentors/2016/09/msg00075.html and https://lists.debian.org/debian-mentors/2016/09/msg00084.html):
MIPS (at least 32bit) doesn't support 64bit atomic operations intrinsically (_8 == 8 bytes). However, gcc provides an emulation library called libatomic. This patch checks for the availability of libatomic and adds it to the linker flags.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -133,7 +133,8 @@ libusbguard_la_LIBADD=\
 	@protobuf_LIBS@ \
 	@udev_LIBS@ \
 	@crypto_LIBS@ \
-	@pegtl_LIBS@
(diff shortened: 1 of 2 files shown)


dkopecek:
Thanks a lot! I've read the posts and I guess that this is the best solution.

PR #140, opened 2016-09-19
Fix Gentoo overlay url

zaolin:
The stuge overlay is deprecated and replaced with my own overlay. Should be officially available in few weeks


dkopecek:
Thanks!

PR #163, opened 2017-02-15
FIX duplicated DBus interface org.usbguard

zezadas:
It seems that DBusInterface.xml has a duplicate interface org.usbguard, and consecutively usbguard-dbus fails checking on expected interfaces number.


dkopecek:
LGTM

PR #170, opened 2017-05-04
Added zsh autocompletion to usbguard

fnoris:
Hi, thank you for your awesome work, I love usbguard! I was about to write my own utility when I found out your code, you saved me time (and I'm sure you did a better job too :) )! I wrote a zsh autocompletion script based on your help menu, I hope it can be useful. Thanks again!

### DIFF
--- a/scripts/usbguard-zsh-completion
+++ b/scripts/usbguard-zsh-completion
@@ -1,0 +1,182 @@
+#compdef usbguard
+#autoload
+
+# usbguard completion, based on cli help text
+
+local curcontext="$curcontext" state line ret=1
+local -a _command_args
+local -a _global_args
+local -a _subcommands
+
+local IFS=$'\n'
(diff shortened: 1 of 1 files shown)


dkopecek:
Hi and many thanks for the contribution! How should the completion file be installed in the system so that zsh picks it up? Is there something similar to the bash-completions package and it's `bash-completion.pc` file:
```
$ pkg-config bash-completion --variable=completionsdir
/usr/share/bash-completion/completions
```

fnoris:
Hi, thank you! Honestly, I don't know how different distros handle the installation properly, sorry. As far as I know, there is a default fpath hardcoded in the zsh binary, /usr/local/share/zsh/site-functions, that should work everywhere. That said, I use Fedora and even though /usr/local/share/zsh/site-functions is actually present in $fpath, Fedora puts everything in /usr/share/zsh/site-functions
Whatever the path, the right file name should be _usbguard (underscore followed by program name)

Issue #160, opened 2017-02-03
usbguard-applet-qt: multiple instances


cgzones:
The usbguard-applet-qt seems not to exit neat on logouts, cause after a following login, there exist multiple instances of it (also multiple icon in the tray):
```
christian@debainl:~$ ps -ef | grep usbguard
root      2592     1  0 16:26 ?        00:00:00 /usr/sbin/usbguard-dbus --system
root      2593     1  0 16:26 ?        00:00:00 /usr/sbin/usbguard-daemon -k -c /etc/usbguard/usbguard-daemon.conf
christi+  3769  3664  0 16:36 ?        00:00:00 /usr/bin/usbguard-applet-qt -session 10cdd3ce61000148613314000000012150010_1486136185_360427
christi+  3783  3664  0 16:36 ?        00:00:00 /usr/bin/usbguard-applet-qt -session 10cdd3ce61000148613580500000028640008_1486136185_362678
christi+  3862  3700  0 16:36 ?        00:00:00 /usr/bin/usbguard-applet-qt
christi+  4289  3824  0 16:36 pts/0    00:00:00 grep usbguard
```

The ksmserver saves the old instances:
```
christian@debainl:~$ cat .config/ksmserverrc 
[General]
screenCount=1

[LegacySession: saved at previous logout]
count=0

[Session: saved at previous logout]
clientId1=10cdd3ce61000147826113500000014890000
clientId2=10cdd3ce61000148613580500000028640001
clientId3=10cdd3ce61000146799440500000031370007
clientId4=10cdd3ce61000147731528400000015010007
clientId5=10cdd3ce61000148613314000000012150010
clientId6=10cdd3ce61000148613580500000028640008
count=6
discardCommand1[$e]=rm,$HOME/.config/session/kwin_10cdd3ce61000147826113500000014890000_1486136185_425449
program1=kwin_x11
program2=/usr/lib/x86_64-linux-gnu/libexec/kdeconnectd
program3=/usr/bin/korgac
program4=/usr/lib/x86_64-linux-gnu/libexec/org_kde_powerdevil
program5=/usr/bin/usbguard-applet-qt
program6=/usr/bin/usbguard-applet-qt
restartCommand1=kwin_x11,-session,10cdd3ce61000147826113500000014890000_1486136185_425449
restartCommand2=/usr/lib/x86_64-linux-gnu/libexec/kdeconnectd,-session,10cdd3ce61000148613580500000028640001_1486136185_362524
[...]

PR #174, opened 2017-05-21
Qt Applet: disable session management

pinotree:
The Qt applet has an autostart `.desktop` file, so it is started at each login of XDG-based environments; since the default session management in Qt applications is to save the application as part of the session,
this leads more instances of the Qt applet to run for each new session, in case the session configuration is to save & restore it.

The solution is to disable the session management for the Qt applet, so it will not be saved by the session manager.

Closes: #160


dkopecek:
Hi.Thanks a lot for this!

PR #179, opened 2017-06-19
Using Policy class

radosroka:
First implementation of policy interface. Just for "Local" source of rules.

### DIFF
--- a/src/Library/Policy.hpp
+++ b/src/Library/Policy.hpp
@@ -20,9 +20,14 @@
 
 #include <Typedefs.hpp>
 #include <string>
+#include <vector>
+
+#include <RuleSet.hpp>
 
 namespace usbguard
 {
+  class Interface;
+
   class DLL_PUBLIC Policy
   {
     public:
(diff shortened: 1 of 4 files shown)


src/Library/Policy.hpp
```diff
@@ -20,19 +20,63 @@
...
     public:
       enum class EventType
       {
         Insert = 1,
         Update = 2,
-        Remove = 3,
+        Remove = 3
+      };
+
+      enum class SourceType
```
dkopecek: Please remove this enum and make the Policy interface class take an instance of RuleSet instead. For example via a method names setRuleSet (in case there will be only one) or addRuleSet / appendRuleSet.

So that the usage is something like:


RuleSet ruleset;
ruleset.load(file);

Policy policy;
policy.setRuleSet(ruleset);
...
...

That way the Policy interface doesn't have to care about the source of the rule set.

src/Library/Policy.hpp
```diff
@@ -20,19 +20,63 @@
...
+        Ldap = 2,
+        SSSD = 3
       };
 
+      Policy(Interface * const interface_ptr);
+
+      const RuleSet getRuleSet();
+      void load();
+      void setPolicySource(SourceType source);
+      void setLocalSourcePath(const String& path);
```
dkopecek: This also doesn't belong in the Policy interface. See my comments above.

radosroka:
I'm closing this PR. I will make another one on feature branch.

PR #180, opened 2017-06-19
usbguard-daemon logs to console by default now

radosroka:
- changed default "-k" bahavior
- option "-k" is disabling logging to console


dkopecek:
Hi, thanks for the PR. We can't just change the behavior of the -k option because that would break existing usage of this option. Instead, I would suggest to leave the -k option as is and add a new option for disabling the console output (-K).

I'm ok with the change of the default value of whether to log to console or not. Should have been the default since the beggining I guess :-)

radosroka:
I have force pushed new commit. It should be OK now.

PR #182, opened 2017-06-20
Using Policy interface

radosroka:


src/Daemon/Daemon.cpp
```diff
@@ -253,7 +252,9 @@ namespace usbguard
   void Daemon::loadRules(const String& path)
   {
     USBGUARD_LOG(Info) << "Loading permanent policy file " << path;
-    _ruleset.load(path);
+    std::shared_ptr<RuleSet> ptr(new RuleSet(this));
+    _policy.setRuleSet(ptr);
+    _policy.load(path);
```
radosroka: This is OK for now. It will be changed in next PR, because I will reduce policy interface later. This is for compatibility reasons only.

src/Library/Policy.hpp
```diff
@@ -20,19 +20,51 @@
 
 #include <Typedefs.hpp>
 #include <string>
+#include <vector>
+
+#include <RuleSet.hpp>
 
 namespace usbguard
 {
+  class Interface;
```
dkopecek: Why this forward declaration?

radosroka: There is no reason for it now.

PR #183, opened 2017-06-21
NSHandler implementation

radosroka:
First simple implementation of NS Handler.

### DIFF
--- a/src/Daemon/Daemon.cpp
+++ b/src/Daemon/Daemon.cpp
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2015 Red Hat, Inc.
+// Copyright (C) 2017 Red Hat, Inc.
 //
 // This program is free software; you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -15,11 +15,13 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 // Authors: Daniel Kopecek <dkopecek@redhat.com>
+//          Radovan Sroka <rsroka@redhat.com>
 //
 #include <build-config.h>
 
 #include "Daemon.hpp"
 #include "Logger.hpp"
+#include "NSHandler.hpp"
 #include "Common/Utility.hpp"
 #include "IPCPrivate.hpp"
 #include "RulePrivate.hpp"
@@ -120,8 +122,12 @@ namespace usbguard
 
   void Daemon::loadConfiguration(const String& path)
   {
+
     USBGUARD_LOG(Info) << "Loading configuration from " << path;
     _config.open(path);
+
+    USBGUARD_LOG(Info) << "Loading NSSwitch...";
+    _nss.parseNSSwitch();
 
     /* RuleFile */
     if (_config.hasSettingValue("RuleFile")) {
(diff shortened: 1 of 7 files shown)


src/Daemon/Daemon.cpp
```diff
@@ -490,14 +496,14 @@ namespace usbguard
...
-      _policy.save(_config.getSettingValue("RuleFile"));
+      _policy.getRuleSet()->save(_config.getSettingValue("RuleFile"));
     }
   }
 
   const RuleSet Daemon::listRules(const std::string& query)
   {
     USBGUARD_LOG(Trace) << "entry: query=" << query; 
-    return _policy.getRuleSet();
+    return (*_policy.getRuleSet());
```
radosroka: I will have to change listRules interface(return value should be shared_ptr). It is not so simple because it is affecting IPC code also.

dkopecek: Yes, this will have to be changed to a pointer type since the RuleSet will become a base class for the various RuleSet implementations (FileRuleSet, LDAPRuleSet, ...)

src/Daemon/Daemon.cpp
```diff
@@ -252,9 +258,10 @@ namespace usbguard
   void Daemon::loadRules(const String& path)
   {
     USBGUARD_LOG(Info) << "Loading permanent policy file " << path;
-    std::shared_ptr<RuleSet> ptr(new RuleSet(this));
-    _policy.setRuleSet(ptr);
-    _policy.load(path);
+
+    auto ruleset = _nss.getRuleSet(this);
+    _policy.setRuleSet(ruleset);
+    ruleset->load(path);
```
dkopecek: From a design perpective I would load the ruleset first, then call setRuleSet, as the load could throw an exception.

src/Library/NSHandler.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (parsed == "") {
+      USBGUARD_LOG(Info) << "There is no \"" << _prop_name << "\"" << " in \"" << _nsswitch_path << "\"";
+      USBGUARD_LOG(Info) << "Using default local source of rules!";
+     } else {
+      USBGUARD_LOG(Info) << "NSSwitch has been parsed. Parsed value is " + parsed;
+     }
+
+    int found = 0;
```
dkopecek: A boolean type would be more appropriate here.

src/Library/NSHandler.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <string>
+#include <vector>
+
+#include "RuleSet.hpp"
+
+#include <pegtl.hh>
+
+namespace usbguard {
+
+  namespace nsswitch {
```
dkopecek: nsswitch seems like an ad-hoc, non-descriptive namespace here. What about NSSwitchParser?

src/Daemon/Daemon.cpp
```diff
@@ -120,9 +122,13 @@ namespace usbguard
 
   void Daemon::loadConfiguration(const String& path)
   {
+
     USBGUARD_LOG(Info) << "Loading configuration from " << path;
     _config.open(path);
 
+    USBGUARD_LOG(Info) << "Loading NSSwich...";
```
dkopecek: Typo: NSSwich => NSSwitch

Issue #30, opened 2015-05-06
Allow to insert new devices only when the user authenticates himself


dkopecek:
This is closely related to the "condition" feature tracked in a separate ticket #24. One use case of such a feature would be to implement a condition that matches an authentication state -- either some sort of a global one and/or user-specific(?). Using that, one could write a policy that allows to connect devices only if the user authenticates himself (password, signature -- smart card device connected, etc.)

dkopecek:
How it could look like when written as a rule given that "if" is used as the keyword for specifying the set of conditions.

```
allow if pki-authenticated(0f3243f0ac)
```
or with multiple conditions
```
allow if one-of { pki-authenticated(keyid) password-authenticated(username) }
```

dkopecek:
This could be implemented easily if usbguard-dbus integrates with polkit.

PR #186, opened 2017-06-29
WIP : rework of the GUI

BenjaminSchubert:
This is currently a work in progress. Please do not merge !

I started by moving the notification system to DBus, to get used to how DBus works with QT.

@dkopecek if you have time, could you look at the code and tell me if that suites you ? So that I don't got too far from what you would like.

This is what the notifications look like : 

On Gnome : 
screenshot from 2017-06-29 16-08-43

On KDE : 
screenshot from 2017-06-29 15-03-23

On Xfce : 
screenshot from 2017-06-29 14-47-07

What do you think about this ?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -569,7 +569,8 @@ CLEANFILES +=\
 	$(top_builddir)/src/GUI.Qt/usbguard.qrc \
 	$(top_builddir)/src/GUI.Qt/DeviceModel.moc.cpp \
 	$(top_builddir)/src/GUI.Qt/SessionBlocker.moc.cpp \
-	$(top_builddir)/src/GUI.Qt/TargetDelegate.moc.cpp
+	$(top_builddir)/src/GUI.Qt/TargetDelegate.moc.cpp \
+	$(top_builddir)/src/GUI.Qt/notifications/DBusNotifier.moc.cpp
 
 nodist_usbguard_applet_qt_SOURCES=\
 	$(usbguard_applet_qt_SOURCES_BASE) \
@@ -578,7 +579,8 @@ nodist_usbguard_applet_qt_SOURCES=\
 	src/GUI.Qt/usbguard.qrc.cpp \
 	src/GUI.Qt/DeviceModel.moc.cpp \
 	src/GUI.Qt/SessionBlocker.moc.cpp \
-	src/GUI.Qt/TargetDelegate.moc.cpp
+	src/GUI.Qt/TargetDelegate.moc.cpp \
+    src/GUI.Qt/notifications/DBusNotifier.moc.cpp
 
 usbguard_applet_qt_LDADD=\
 	$(top_builddir)/libusbguard.la \
--- a/Makefile.am
+++ b/Makefile.am
@@ -530,7 +530,10 @@ usbguard_applet_qt_SOURCES_BASE=\
 	src/GUI.Qt/SessionBlocker.cpp \
 	src/GUI.Qt/SessionBlocker.h \
 	src/GUI.Qt/TargetDelegate.cpp \
-	src/GUI.Qt/TargetDelegate.h
+	src/GUI.Qt/TargetDelegate.h \
+	src/GUI.Qt/notifications/AbstractNotifier.h \
+	src/GUI.Qt/notifications/DBusNotifier.cpp \
+	src/GUI.Qt/notifications/DbusNotifier.h
 
 EXTRA_DIST+=\
 	src/GUI.Qt/DeviceDialog.ui \
--- a/Makefile.am
+++ b/Makefile.am
@@ -531,9 +531,12 @@ usbguard_applet_qt_SOURCES_BASE=\
 	src/GUI.Qt/SessionBlocker.h \
 	src/GUI.Qt/TargetDelegate.cpp \
 	src/GUI.Qt/TargetDelegate.h \
+	GUI.Qt/backend/AbstractBackend.h \
+	src/GUI.Qt/backend/DBusBackend.cpp \
+	src/GUI.Qt/backend/DBusBackend.h \
 	src/GUI.Qt/notifications/AbstractNotifier.h \
 	src/GUI.Qt/notifications/DBusNotifier.cpp \
-	src/GUI.Qt/notifications/DbusNotifier.h
+	src/GUI.Qt/notifications/DBusNotifier.h
 
 EXTRA_DIST+=\
 	src/GUI.Qt/DeviceDialog.ui \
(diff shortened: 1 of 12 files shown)


dkopecek:
Hello and thanks! The code looks really good. I don't think the style diverges in any significant way. I would even say the overall quality is better than rest of the code base since you are disciplined and include even the doxygen docs :-)

dkopecek:
btw. did you abandon the idea of writing the applet from scratch?

Makefile.am
```diff
@@ -530,7 +530,10 @@ usbguard_applet_qt_SOURCES_BASE=\
 	src/GUI.Qt/SessionBlocker.cpp \
 	src/GUI.Qt/SessionBlocker.h \
 	src/GUI.Qt/TargetDelegate.cpp \
-	src/GUI.Qt/TargetDelegate.h
+	src/GUI.Qt/TargetDelegate.h \
+	src/GUI.Qt/notifications/AbstractNotifier.h \
+	src/GUI.Qt/notifications/DBusNotifier.cpp \
+	src/GUI.Qt/notifications/DbusNotifier.h
```
dkopecek: Typo: DbusNotifier.h => DBusNotifier.h

BenjaminSchubert:
@dkopecek I've removed completely the use of the IPC backend from the system.

As of now, the DBus backend client is in Gui.Qt/backend, where I also specified an AbstractBackend. Would it make sense to make IPCClient inheritate of the backend so that we have a common interface between the two and we could therefore easily swap one for the other ? If so, should we extract the common backend to another package ? What do you think ?

Concerning Polkit, I am wheel on my system with the default polkit rules for Fedora. However, I don't get to enter my password when changing rules, I can do it automatically, is this normal ? 

Otherwise, I think a first merge would be doable once you find the code acceptable. I have tested it locally but i may have introduced some bugs... maybe creating a new branch for the GUI updates and pushing all the updates I/we are doing to do in it before merging to master would be better ? What do you think ? If you create a new branch, I could modify this PR to be on the new branch so we can centralize the changes better.

dkopecek:
> As of now, the DBus backend client is in Gui.Qt/backend, where I also specified an AbstractBackend. Would it make sense to make IPCClient inheritate of the backend so that we have a common interface between the two and we could therefore easily swap one for the other ? If so, should we extract the common backend to another package ? What do you think ?

Well, there's already the Interface class which defines the common interface. Is that not usable? We can't make IPCClient depend on the AbstractBackend as that includes the Qt QObject system and that's a dependency I really don't want to see in the library. If you drop the QObject stuff from the backend, then what would be the difference from the Interface class?

> Concerning Polkit, I am wheel on my system with the default polkit rules for Fedora. However, I don't get to enter my password when changing rules, I can do it automatically, is this normal ? 

That seems to be because of the missing polkit auth check you mentioned in the e-mail. I've never really tested the D-Bus bridge as I've never got to rewriting the applet to use D-Bus. Now there seems to be interest in the desktop integration so it makes sense to get this part of the project in shape.

> Otherwise, I think a first merge would be doable once you find the code acceptable. I have tested it locally but i may have introduced some bugs... maybe creating a new branch for the GUI updates and pushing all the updates I/we are doing to do in it before merging to master would be better ? What do you think ? If you create a new branch, I could modify this PR to be on the new branch so we can centralize the changes better.

Ok, that seems reasonable. What about a "feature-gui-rework" branch?

> Also there is quite a lot of mixes between the std library and the Qt one, what should we promote ? Should the UI use only Qt, which seems cleaner ? If we separate the backend, should we remove all dependencies on Qt ? What do you prefer ?

Yeah, the GUI should preferably use Qt.

> Concerning the naming, should I name my header files as .h or .hpp ? I saw both used in the project.

I prefer .hpp but I had some problems with that when dealing with the Qt code via Qt creator. However, if there is no problem now with using custom extenstions (i.e. ".h" is not hardcoded/assumed at any point in the Qt toolchain or Qt creator), then prefer .hpp.

BenjaminSchubert:
> Well, there's already the Interface class which defines the common interface. Is that not usable? We can't make IPCClient depend on the AbstractBackend as that includes the Qt QObject system and that's a dependency I really don't want to see in the library. If you drop the QObject stuff from the backend, then what would be the difference from the Interface class?

Wow... I missed that completely. I will refactor my code to use the Interface.

> That seems to be because of the missing polkit auth check you mentioned in the e-mail. I've never really tested the D-Bus bridge as I've never got to rewriting the applet to use D-Bus. Now there seems to be interest in the desktop integration so it makes sense to get this part of the project in shape.

Yes, I will work on this in parallel.


> Ok, that seems reasonable. What about a "feature-gui-rework" branch?
perfect !

I will not have much times the next few weeks but I'll try to do my best !

MatejKovacic:
A short suggestion for how GUI should look like.

There should be the following options (available on right click):
- Show rules (which would also allow to edit rules and create default policy)
- Show events (which will show how and when USB devices were connected and blocked/allowed; maybe events would be available just for a short period)
- Settings (autostart, allowed users,...)
- Quit (which should also stop the USBGuard system service, before that there should be a confirmation dialog with warning, that USBGuard will now be inactive)

For notifications please see #203. Basically the notifications should be:
- at first connection/if there is no rule for deice: "USB device inserted. Device is temporarily blocked. \nUSB ID, name, port)".
- when user defines rule for a device: "New rule set for device. Device is blocked/allowed. \nUSB ID, name, port".
- after a rule for a device already exists: "USB device inserted. Device is blocked/allowed. \nUSB ID, name, port".

Notification should expire as soon as user does his action (click block or allow) or when there is new USBGuard event (for instance, someone quickly inserts mouse and USB key).

User dialog for new rule now looks terrible. :-)

My suggestion is it displays the following text:

Big text: New USB device inserted.
Device name: ***
(smaller text): Device ID: ***, serial number: *** .

USBGuard detected new USB device for which there is no rule yet. Would you like to allow or block the device?

(Rounded buttons): Allow, Block.

(Tick 1): Create permanent rule.
(Tick 2): Allow USB device on this port only.

Something like that. Probably the UX could be improved.

MatejKovacic:
I have another idea. There should be option that user with sudo privileges would need to enter his password to allow using the device. This would prevent malicious user to set the rule or allow the USB device when legitimate user is not present (i.e. computer is left unattended). 

BTW, is there any time plan when this new GUI is going to be implemented?

dkopecek:
@MatejKovacic The user authentication feature is tracked/requested also in #30. It is implementable by using the dbus bridge (usbguard-dbus) and if PolKit auth check gets added there. @BenjaminSchubert already knows about the missing check and will probably add it as part of this GUI rework, if he has time to continue to work on this PR.

BenjaminSchubert:
Yeah sorry, I'm currently very busy and might be for the a month still.

If someone wants to go with the missing check on the backend, it will be with pleasure, otherwise I'll work on it once I can have time for the GUI rework !

dkopecek:
Closing as this is inactive for too long.

PR #192, opened 2017-07-13
Implemented policy file permission check

tammar96:
I added permission check to policy files. It can be avoided
by using switch -P.

### DIFF
--- a/src/Daemon/Daemon.cpp
+++ b/src/Daemon/Daemon.cpp
@@ -119,16 +119,60 @@ namespace usbguard
     _config.close();
   }
 
-  void Daemon::loadConfiguration(const std::string& path)
+  int Daemon::checkPermissions(const std::string& path,
+                               const mode_t permissions)
+  {
+    struct stat file_stat;
+
+    // from all enabled permissions we subtract the permissions we want to check
+    // after this operation variable permission_bad contains the complement of 
+    // permissions we want to check.
+    mode_t permission_bad { ( S_IRWXU | S_IRWXG | S_IRWXO ) - permissions };
+
+    if(!stat(path.c_str(), &file_stat)) {
+      if (S_ISREG(file_stat.st_mode)) {
+
+        // this comparison inspect if file has the wanted permissions and if
+        // the file does not contain the unwanted permissions.
+        if (  !( file_stat.st_mode & permissions ) ||
+               ( file_stat.st_mode & permission_bad)
+           )
+        {
+          USBGUARD_LOG(Error) << "Policy leaks are possible with this permissions!";
+          throw Exception("Check permissions", path, "Policy leaks possible");
+        }
+        else {
+          USBGUARD_LOG(Info) << "File has correct permissions.";
+        }
+      }
+      else {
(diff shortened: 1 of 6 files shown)


src/Tests/UseCase/001_cli_policy.sh
```diff
@@ -40,17 +40,17 @@ function test_cli_policy()
 {
   set -e
   sleep 4
-  
+
```
dkopecek: whitespace only changes, please get rid of this.

src/Daemon/main.cpp
```diff
@@ -96,6 +98,9 @@ int main(int argc, char *argv[])
       case 'p':
 	pid_file = std::string(optarg);
 	break;
+      case 'P':
+  USBGUARD_LOG(Warning) << "PERMISSIONS CHECK ON POLICY FILE ARE TURNED OFF!", check_permissions = false;
```
dkopecek: Please don't use the ',' operator to separate the expressions. Just write it on 2 lines. IMHO it's more readable that way.

src/Daemon/Daemon.cpp
```diff
@@ -119,16 +119,53 @@ namespace usbguard
...
+        }
+      }
+      else {
+        USBGUARD_LOG(Error) << "ERROR: In specified path is not located file.";
+        throw Exception("Check permissions", path, "File not exists");
+      }
+    }
+    else {
+      USBGUARD_LOG(Error) << "ERROR: obtaining file permissions!";
+      throw Exception("Check permissions", path, "Get file permissions");
```
dkopecek: It would be better to use the ErrnoException here (see https://github.com/dkopecek/usbguard/blob/master/src/Library/public/usbguard/Exception.hpp#L112):
```
throw ErrnoException("Check permissions", path, errno);
```

Also include the errno value in the log output.

src/Daemon/Daemon.cpp
```diff
@@ -119,16 +119,53 @@ namespace usbguard
...
+        {
+          USBGUARD_LOG(Error) << "Policy leaks are possible with this permissions!";
+          throw Exception("Check permissions", path, "Policy leaks possible");
+        }
+        else {
+          USBGUARD_LOG(Info) << "File has correct permissions.";
+        }
+      }
+      else {
+        USBGUARD_LOG(Error) << "ERROR: In specified path is not located file.";
```
dkopecek: I don't understand the error message. S_ISREG == false doesn't mean that there's not file at the location, it means that there is one but it's not a regular file. Please provide better wording based on the check that you do on the file.

src/Daemon/Daemon.cpp
```diff
@@ -119,16 +119,53 @@ namespace usbguard
     _config.close();
   }
 
-  void Daemon::loadConfiguration(const std::string& path)
+  int Daemon::checkPermissions(const std::string& path,
+                               const mode_t permissions)
+  {
+    struct stat file_stat;
+    mode_t permission_bad { ( S_IRWXU | S_IRWXG | S_IRWXO ) - permissions };
```
dkopecek: Please explain in a comment in the code why we need this set of hard coded permission flags and the logic of the check.

PR #194, opened 2017-07-17
RuleSet refactoring

radosroka:
- created RuleSet abstract class
- created FileRuleSet derived class for local policy
- started using shared_ptr<RuleSet> instead of const RuleSet&


dkopecek:
Please rename RuleSetLocal to FileRuleSet. "Local" seems abstract to me. Also, it doesn't pass the use case tests, please look into it.

PR #206, opened 2017-09-26
Basic LDAP support

radosroka:


src/Library/public/usbguard/Rule.cpp
```diff
@@ -243,14 +245,20 @@ namespace usbguard {
...
-	     getTarget() == Target::Invalid);
+        getTarget() == Target::Invalid);
   }
 
   std::string Rule::toString(bool invalid) const
   {
     return d_pointer->toString(invalid);
   }
 
+  std::string Rule::toLDIF(std::string hostname, std::string name_prefix, std::string USBGUARD_BASE, std::string objclass,
```
dkopecek: I don't like that you are introducing LDAP specifics into the Rule interface. This has to go away.

And btw. why passing by value and not reference?

PR #210, opened 2017-11-23
KeyValueParser update

tammar96:
Updated configParser

### DIFF
--- a/src/Library/KeyValueParser.cpp
+++ b/src/Library/KeyValueParser.cpp
@@ -1,0 +1,141 @@
+//
+// Copyright (C) 2015 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+// Authors: Daniel Kopecek <dkopecek@redhat.com>
+// Authors: Marek Tamaskovic <mtamasko@redhat.com>
+//
+#ifdef HAVE_BUILD_CONFIG_H
+  #include <build-config.h>
+#endif
+
+#include "KeyValueParser.hpp"
+#include "Common/Utility.hpp"
+
+#include "usbguard/Exception.hpp"
+#include "usbguard/Logger.hpp"
+
+namespace usbguard
+{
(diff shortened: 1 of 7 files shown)


src/Library/KeyValueParser.cpp
```diff
@@ -0,0 +1,141 @@
...
+        return false;
+      }
+    }
+
+    return 1;
+  }
+
+  bool KeyValueParser::checkMapValidity()
+  {
+    USBGUARD_LOG(Error) << "ERROR: Override KeyValueParser::checkMapValidity() method!" << std::endl;
```
dkopecek: You should declare checkMapValidity as a pure virtual method instead of polluting the log at run time. If there has to be an implementation of that function, then why not force it at compile time...

src/Library/KeyValueParser.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  bool KeyValueParser::checkKeyValidity(const std::string& key)
+  {
+    for (auto a: this->keys) {
+      if (!key.compare(a)) {
+        return false;
+      }
+    }
+
+    return 1;
```
dkopecek: 1 => true

src/Library/KeyValueParser.hpp
```diff
@@ -0,0 +1,60 @@
...
+    parsed_t(const std::string a, const std::string b): lvalue(a), rvalue(b) {};
+  };
+
+  class KeyValueParser
+  {
+    std::vector<std::string> keys;
+    std::string separator {""};
+    std::map<std::string, std::string> output_map;
+    bool checkKeyValidity(const std::string& key);
+    virtual bool checkMapValidity();
```
dkopecek: What is this method supposed to do? Checking the whole map after it is parsed doesn't seem to be a good solution as the information about the source of an error is lost at that point (information like line or column number).

src/Library/KeyValueParser.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <memory>
+
+namespace usbguard
+{
+
+  struct parsed_t {
+    const std::string lvalue;
+    const std::string rvalue;
+    parsed_t(const std::string a, const std::string b): lvalue(a), rvalue(b) {};
+  };
```
dkopecek: Why do you use your own type instead of std::pair?

src/Library/KeyValueParser.hpp
```diff
@@ -0,0 +1,60 @@
...
+    std::string separator {""};
+    std::map<std::string, std::string> output_map;
+    bool checkKeyValidity(const std::string& key);
+    virtual bool checkMapValidity();
+
+  public:
+    KeyValueParser(const std::vector<std::string>& v);
+    KeyValueParser(const std::vector<std::string>& v, const std::string& sep);
+    std::unique_ptr<struct parsed_t> parseString(std::string& str);
+    bool parseStream(std::fstream& stream);
```
dkopecek: Why does parseStream return bool and parseString returns parsed_t? There cannot be multiple key=value pairs in a string passed to parseString?

PR #238, opened 2018-07-09
Add alternative to stdio_filebuf.h for libc++ support.

Allen-Webb:
I had to add the .m4 files to find boost to get this to work, but it uses boost iostreams to replace the functionality of ext/stdio_filebuf.h that isn't present in libc++.


Allen-Webb:
I had to set the --with-boost-libdir=<boost parent dir> flag when running configure to get it to build when ext/stdio_filebuf.h wasn't present. I am not that familiar with the LGTM analysis tool, so I don't know what changes need to be made since this adds a new dependency if ext/stdio_filebuf.h is not found.

Allen-Webb:
I fixed the problem, and fixed the code style which was causing the checks to fail.

PR #244, opened 2018-07-27
Added fuzzer targets for oss-fuzz.

Allen-Webb:
This commit needs to be merged before I can merge this:
https://github.com/Allen-Webb/oss-fuzz/tree/usbguard


dkopecek:
This is great, thanks a lot!

PR #249, opened 2018-09-13
Namespace generically named LDAP attributes

grawity:
I noticed that recently LDAP support was added to USBGuard, which adds some attributes including the generically named `RuleTarget` and `RuleOrder`.

Since attribute names have to be unique across the LDAP server, even between schemas (and no, the OIDs are mostly just for show), this has somewhat greater chance of colliding with someone's existing schema in the future.

So while the code is still warm, I would recommend prefixing those two attributes with `USBGuard…` or at least `USB…`, like the remaining ones. With namespacing, name collisions are much less likely.

### DIFF
--- a/scripts/ldap/schema2ldif.pl
+++ b/scripts/ldap/schema2ldif.pl
@@ -1,0 +1,50 @@
+#!/usr/bin/env perl
+# Converts OpenLDAP schema from traditional slapd.conf format to LDIF format
+# usable for importing into cn=config.
+#
+# Copyright (c) 2012-2016 Mantas Mikulėnas <grawity@gmail.com>
+# Released under the MIT license <https://spdx.org/licenses/MIT>
+
+use warnings;
+use strict;
+
+my $name = shift(@ARGV) // "UNNAMED-SCHEMA";
+my $unwrap = 0;
+
+print "dn: cn=$name,cn=schema,cn=config\n";
+print "objectClass: olcSchemaConfig\n";
+
+my $key;
+my $value;
+
+while (<STDIN>) {
+	if (/^(attributeType(?:s)?|objectClass(?:es)?) (.+)$/i) {
+		if ($key && $value) {
+			print "$key: $value\n";
+		}
+		($key, $value) = ($1, $2);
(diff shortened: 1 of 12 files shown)


grawity:
I noticed that there already is a `USBGuardOrder` attribute, which the source code doesn't use for anything. Was that supposed to be the same as `RuleOrder`?

radosroka:
It seems that USBGuardOrder was first and then there was an intention to rename it as RuleOrder. It is so for src/{Tests|Daemon|Common}/* files and scripts/ldap/* files seem to be not updated.

I think that it would be nice to have scripts/ldap/* files as primary location for ldap scripts and then we can have symlinks in tests to these scripts. Because now we have these scripts on two places and even not updated.

radosroka:
```
 std::vector<std::string> LDAPUtil::_rule_keys = {
    "RuleTarget", /* just for indexing */
    "USBGuardHost", /* just for indexing */
    "USBGuardOrder", /* just for indexing */
    "id",
    "serial",
```
In this case USBGuardOrder is just placeholder, there is no Order in regular (file based) rule.

scripts/ldap/schema2ldif.pl
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env perl
```
dkopecek: Please add a copyright header to this file.

grawity: As in `Copyright (c) 2018 Red Hat, Inc.`?

dkopecek: I'm not really an expert either, but the primary reason is to have each contributed file licensed properly and a reference to it's author(s). Another reason is that for example the Debian Packaging Guidelines are very strict about the licensing stuff and require (AFAIK) that every project file has a license header.

So if you are the author of the script, please record that in the header. Choose whatever OSI compatible license you wish. The easiest route is to go with what USBGuard uses itself.

grawity: Since the original is already available with a MIT license header in other locations, after some research I decided to keep that here as well.

dkopecek:
Thanks!

PR #250, opened 2018-09-14
Add a daemon config option HidePii.

Allen-Webb:
This new configuration option excludes USB device serial numbers and
the devices hashes (which include the serial number) from being
logged to the Audit backend. This is useful when you want the audit
entries without PII to be writen to the syslog.


Allen-Webb:
I made the capitalization change.

dkopecek:
Thanks!

PR #252, opened 2018-09-17
Add devpath option to generate-policy.

Allen-Webb:
This allows usbguard generate-policy to generate a policy line for a single device given the devpath without forcing a reenumeration of all the devices.

### DIFF
--- a/src/Library/UEventDeviceManager.cpp
+++ b/src/Library/UEventDeviceManager.cpp
@@ -316,6 +316,27 @@ namespace usbguard
     }
   }
 
+  void UEventDeviceManager::scan(const std::string& devpath)
+  {
+    std::vector<std::string> components;
+    tokenizeString(devpath, components, "/", /*trim_empty=*/true);
+    auto base = std::find_if(components.begin(), components.end(),
+    [](const std::string& component) -> bool {
+      return hasPrefix(component, "usb");
+    });
+    std::string path = "";
+
+    for (auto itr = components.begin(); itr < components.end(); ++itr) {
+      path += "/" + *itr;
+
+      if (itr >= base) {
+        ueventProcessAction("add", path);
+      }
+    }
+
+    ueventProcessAction("add", path);
+  }
+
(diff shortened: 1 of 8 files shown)


src/Library/UEventDeviceManager.cpp
```diff
@@ -316,6 +316,90 @@ namespace usbguard
...
+      DeviceException(ex.message());
+      return;
+    }
+    catch (...) {
+      USBGUARD_LOG(Warning) << "USB Device Exception: unknown exception";
+      DeviceException("unknown exception");
+      return;
+    }
+  }
+
```
dkopecek: This is a big code duplication. Would it be possible to refactor this and implement it somehow based on the existing scan method -- i.e. do a bit of refactoring to be able to share the common code?

Allen-Webb: I was able to split ueventProcessUEvent in a way that allowed me to remove a big chunck of the duplicate code while not triggering uevents on the netlink socket.

Allen-Webb:
I found a case where this isn't working as intended so I need to do some troubleshooting before this merges.

Allen-Webb:
I fixed the problem I ran into, but Travis CL is failing with these errors:
...
UseCase/002_cli_devices.sh.log:  modprobe: FATAL: Module dummy_hcd not found.
UseCase/002_cli_devices.sh.log:  +sudo -n rmmod g_mass_storage
UseCase/002_cli_devices.sh.log:  rmmod: ERROR: Module g_mass_storage is not currently loaded
UseCase/002_cli_devices.sh.log:  +sudo -n modprobe g_mass_storage file=/tmp/usbguard_disk iSerialNumber=555666111
UseCase/002_cli_devices.sh.log:  modprobe: ERROR: could not insert 'g_mass_storage': No such device
...
It looks like a test environment setup issue to me (missing kernel modules).

dkopecek:
LGTM. Thanks for investing time into this feature and the subsequent code deduplication!

PR #264, opened 2018-11-07
Made the netlink socket and thread optional for UEventDeviceManager.

Allen-Webb:
Now when invoking usbguard generate-policy -d ${devpath}, no netlink
socket or listener thread is spawned. This removes a race condition
where a device connecting while scan(devpath) was executing might be
included in the policy list.


Allen-Webb:
I am actually going to revise this so it doesn't need root.

dkopecek:
Thanks! I think this makes sense.

PR #274, opened 2019-02-07
Add support for portX/connect_type attribute

gwinkenbach:
Add with-connect-type rule attribute to filter and control devices through the portX/connect_type flag. This flag is getting more commonly properly initialized and is useful, for instance, to distinguish between internal ("hardwired") and external ("hotplug") ports.


dkopecek:
Thank you! This is a quite useful attribute to include in the policy.

PR #281, opened 2019-03-14
Add daemon configuration option for authorized_default.

Allen-Webb:
This is follow up work for commit 4805aba1 pull request #274.

### DIFF
--- a/src/Library/UEventDeviceManager.cpp
+++ b/src/Library/UEventDeviceManager.cpp
@@ -48,6 +48,29 @@
 
 namespace usbguard
 {
+  namespace
+  {
+    void setDeviceAuthorizedDefault(SysFSDevice* device, DeviceManager::AuthorizedDefaultType auth_default)
+    {
+      if (auth_default == DeviceManager::AuthorizedDefaultType::Keep) {
+        return;
+      }
+
+      std::string auth_default_str = std::to_string(DeviceManager::authorizedDefaultTypeToInteger(auth_default));
+      device->setAttribute("authorized_default", auth_default_str);
+
+      if (device->readAttribute("authorized_default", /*strip_last_null=*/true) != auth_default_str) {
+        if (auth_default == DeviceManager::AuthorizedDefaultType::Internal) {
+          USBGUARD_LOG(Warning) << "No kernel support for authorized_default = 2, falling back to 0";
+          setDeviceAuthorizedDefault(device, DeviceManager::AuthorizedDefaultType::None);
+        }
+        else {
+          throw Exception("UEventDevice", device->getPath(), "Failed to set authorized_default to \"" + auth_default_str + "\"");
+        }
+      }
+    }
+  }  /* namespace */
+
   UEventDevice::UEventDevice(UEventDeviceManager& device_manager, SysFSDevice& sysfs_device)
     : Device(device_manager)
   {
@@ -102,11 +125,11 @@ namespace usbguard
        */
       setTarget(Rule::Target::Block);
     }
+
     /*
      * Set connect type
      */
     setConnectType(sysfs_device.readAttribute("port/connect_type", /*strip_last_null=*/true, /*optional=*/true));
-
     /*
      * Process USB descriptor data.
      *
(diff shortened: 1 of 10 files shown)


src/Library/UEventDeviceManager.cpp
```diff
@@ -245,21 +245,40 @@ namespace usbguard
   /*
    * Manager
    */
+  void UEventDeviceManager::setDeviceAuthorizedDefault(SysFSDevice* device, DeviceManager::AuthorizedDefaultType auth_default)
```
Allen-Webb: It might make sense to move this to DeviceManager instead of UEventDeviceManager.

Allen-Webb:
> Hi Allen, thanks for this feature! Could you please remove the .orig files from the PR? Otherwise it seems fine and I'm going to merge it.

Sorry about missing those when I uploaded the commit. It should be clean now. Thanks.

dkopecek:
no problem, thanks!

Cropi:
Hi @Allen-Webb,

This is an amazing feature in the usbguard that I would like to use. However, I am having problems when AuthorizedDefault is set to 'wired'. Can you take a look at it? Thanks.

src/Library/public/usbguard/DeviceManager.hpp
```diff
@@ -44,23 +44,38 @@ namespace usbguard
       Remove = 3,
     };
 
+    enum class AuthorizedDefaultType {
```
Cropi, line 47: According to the [kernel documentation](https://www.kernel.org/doc/html/latest/usb/authorization.html) about usb authorization there are only 3 different values the `authorized_default` file can hold, including:
-  "0" New devices connected to hostX to be deauthorized by default
-  "1" New devices connected to hostX to be authorized by default
-  "2" This causes kernel to only authorize by default devices connected to internal USB ports.

However, when AuthorizedDefault is set to 'wired' in the usbguard-daemon.conf.in file, the daemon will attempt to write a "-1" to the appropriate authorized_default file, which will fail.

Allen-Webb: I am checking to see if this is due to a change in the patch that landed in the CrOS kernel and the one that landed in upstream. If so, I will work to resolve it.

Allen-Webb: It looks like it is a documentation problem. See the source code here:
https://github.com/torvalds/linux/blob/master/drivers/usb/core/hcd.c#L369

Cropi: Could you please verify that AuthorizedDefault=wired works for you?  Thank you in advance .

Allen-Webb: It works for me, but it may not work for people running an older kernel versions which is fine as long as usbguard-daemon falls back to 0.

Cropi: May I ask which version of the kernel(and linux distribution) you are using?

Allen-Webb: I took a look at this again and I was confusing wired(-1) with internal(2). I had done some work earlier to get internal devices added.

My understanding is -1 was present because of wireless USB, but I don't know why setting -1 doesn't work for a typical hub. I am hesitant to remove it without understanding why it is still present in the kernel sources.

Cropi: I agree with you. This particular part of code in the kernel is confusing. According to the source documentation it should be doable, however it's not.

I've tried it on older kernel versions to verify that this is a kernel bug but I had no luck making it to work.

radosroka: After some digging in kernel sources it seems that `wired` was never an option for setting in `authorized_default`. It was implemented as a kernel internal state for enabling wired controllers during boot. The USBGuard daemon is not running at this point so kernel boots up with a limited set of devices. Setting `authorized_default` to `wired` would be useless after boot because only host controller can handle the value and it is already allowed.

Can anybody confirm my findings?

radosroka: @Allen-Webb @Cropi I believe we should remove the `wired` option totally from the USBGuard.

Cropi: Yes, I agree with you.

PR #285, opened 2019-04-09
Add temporary option to append-rule

tweksteen:
Similarly to the -p (permanent) option of allow-device, this commit adds the option to append a rule only temporary. I have kept the name as "permanent" on the service side but exposed as the opposite (temporary) on the CLI to match the current default behavior (that is, appending is currently permanent).

### DIFF
--- a/src/CLI/usbguard-append-rule.cpp
+++ b/src/CLI/usbguard-append-rule.cpp
@@ -29,11 +29,12 @@
 
 namespace usbguard
 {
-  static const char* options_short = "ha:";
+  static const char* options_short = "ha:t";
 
   static const struct ::option options_long[] = {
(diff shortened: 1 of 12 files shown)


tweksteen:
I've tried to reproduce the failing test case locally without success. Any idea why this is failing?

muelli:
very nice change. we're looking forward to using that for temporarily disallowing USB while the screen is locked. 

The D-Bus interface is not versioned, it seems. If it was, it would need a bump.

dkopecek:
> The D-Bus interface is not versioned, it seems. If it was, it would need a bump.

Maybe that's something we could improve before next release. Is it something that can be expressed in the XML interface file? i.e. I don't know how D-Bus versioning works :-)

PR #292, opened 2019-05-27
Remove _default_action attribute from RuleSet

tweksteen:
The _default_action attribute is not currently used and appears to be a functional duplicate of _default_target. Remove it and the associated methods.


dkopecek:
You are exactly right in your comment. The actions were intended to provide a mechanism to trigger custom commands or something like that, but I have never thought it through and implemented it. It would require to implement things that are out of scope for the daemon. Instead, custom actions/commands can be triggered via "usbguard watch -e <script>" which is a much simpler and nicer (IMHO) way to do that. Thanks for cleaning that up!

PR #302, opened 2019-06-17
Add versioning to DBus service

tweksteen:


tweksteen:
@muelli Please take a look

dkopecek:
@tweksteen So, when we make a signature change in e.g. Policy.listRules, what happens then with the numbers? usbguard1 becomes usbguard2 or Policy.listRules gets a new suffix? I've read http://0pointer.de/blog/projects/versioning-dbus.html but I still don't know how I would implement it properly. There are several layers where things can change and probably depends whether we keep the old interfaces/methods there or provide only the new ones under new names...?

* Does changing a method signature bump version of only the method or the parent interface?
* Does adding a method bump version of an interface? (it is still backwards compatible in that case, no?)

muelli:
LGTM



>     * Does changing a method signature bump version of only the method or the parent interface?
> 

I haven't seen anybody doing versioned methods, so I guess "parent interface" is the answer.

>     * Does adding a method bump version of an interface? (it is still backwards compatible in that case, no?)

The way I read http://0pointer.de/blog/projects/versioning-dbus.html I'd say that you don't bump the version number. But you could argue that it makes the lives of clients easier if the number is bumped, though, because they don't have to handle the `UnknownMethod`.

tweksteen:
> I haven't seen anybody doing versioned methods, so I guess "parent interface" is the answer.

I did have a look on my local buses but I couldn't see any service name (nor interface) with version >1. @muelli Any practical example you could share with us?

> The way I read http://0pointer.de/blog/projects/versioning-dbus.html I'd say that you don't bump the version number. But you could argue that it makes the lives of clients easier if the number is bumped, though, because they don't have to handle the UnknownMethod.

That's my understanding as well. No strong opinion here, happy to go either way.

muelli:
> but I couldn't see any service name (nor interface) with version >1. @muelli Any practical example you could share with us?

Evolution bumps DBus API relatively often, e.g. https://gitlab.gnome.org/GNOME/evolution-data-server/commit/78142ce3472b18355b25832e72e2bb36a23c26e2 cf. https://gitlab.gnome.org/GNOME/evolution-data-server/commit/78142ce3472b18355b25832e72e2bb36a23c26e2

Rhythmbox also seems to have bumped DBus API versions (to 3).

src/DBus/DBusBridge.hpp
```diff
@@ -28,6 +28,14 @@
 #include <gio/gio.h>
 #pragma clang diagnostic pop
 
+#define DBUS_BUS_NAME          "org.usbguard1"
+#define DBUS_ROOT_INTERFACE    "org.usbguard1"
+#define DBUS_ROOT_PATH         "/org/usbguard1"
+#define DBUS_POLICY_INTERFACE  "org.usbguard1.Policy"
```
dkopecek: @tweksteen What still confuses me:
```
"org.usbguard1.Policy"
  Why here --^       ^--- and not here?
```

tweksteen: Right, I agree. Re-read the blog post and with @muelli's explanation that make sense.

tweksteen:
Pushed an updated version. Renamed `DBUS_BUS_NAME` to `DBUS_SERVICE_NAME`. Moved the versioning of the interface at the end of their name. I kept the versioning of the object path as `/org/usbguard1/...` as these should be changed only when the service name itself is updated (not the interface). This follows the DBus versioning best practices.

PR #317, opened 2019-07-17
Extend the functionality of allow/block/reject commands to work with rules

Cropi:
These commands are now able to work with either a **device id** or a specific **device rule**.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ before_install:
 - sudo apt-get install -y asciidoc
 - sudo apt-get install -y umockdev libumockdev-dev
 - sudo gem install coveralls-lcov
-- curl -JLo /tmp/astyle.tgz --remote-header-name 'https://netix.dl.sourceforge.net/project/astyle/astyle/astyle%203.1/astyle_3.1_linux.tar.gz'
+- curl -JLo /tmp/astyle.tgz --remote-header-name 'https://netcologne.dl.sourceforge.net/project/astyle/astyle/astyle%203.1/astyle_3.1_linux.tar.gz'
 install:
 - pushd .
 - cd /tmp
--- a/src/CLI/usbguard-allow-device.cpp
+++ b/src/CLI/usbguard-allow-device.cpp
@@ -22,6 +22,8 @@
 
 #include "usbguard.hpp"
 #include "usbguard-allow-device.hpp"
+#include "usbguard/RuleParser.hpp"
+#include "Common/Utility.hpp"
 
 #include "usbguard/IPCClient.hpp"
 
@@ -39,7 +41,7 @@ namespace usbguard
 
   static void showHelp(std::ostream& stream)
   {
-    stream << " Usage: " << usbguard_arg0 << " allow-device [OPTIONS] <device-id>" << std::endl;
+    stream << " Usage: " << usbguard_arg0 << " allow-device [OPTIONS] (<device-id> | <rule>)" << std::endl;
     stream << std::endl;
     stream << " Options:" << std::endl;
     stream << "  -p, --permanent  Make the decision permanent. A device specific allow" << std::endl;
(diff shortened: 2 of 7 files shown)


src/CLI/usbguard-allow-device.cpp
```diff
@@ -75,14 +76,47 @@ namespace usbguard
...
+      }
+      catch (...) {
+        showHelp(std::cerr);
+        return EXIT_FAILURE;
+      }
+
+      /*
+       * Create a ruleset which will contain the requested rule
+      */
+      auto my_rule_set = std::dynamic_pointer_cast<RuleSet>(std::make_shared<MemoryRuleSet>(nullptr));
```
dkopecek: I don't think you have to create a RuleSet here. Wouldn't it be possible to use the [Rule::appliesTo](https://github.com/USBGuard/usbguard/blob/27801e8800b6c2ebf34a6104d318ae541ff66706/src/Library/public/usbguard/Rule.hpp#L490) method? i.e. something like:
```
if (rule.appliesTo(rule_device)) ... 
```

src/CLI/usbguard-allow-device.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "usbguard.hpp"
 #include "usbguard-allow-device.hpp"
+#include "PolicyGenerator.hpp"
```
dkopecek: Including this is unnecessary, I think you want `#include "usbguard/Rule.hpp"`.

src/CLI/usbguard-allow-device.cpp
```diff
@@ -75,14 +76,47 @@ namespace usbguard
...
-    if (argc != 1) {
-      showHelp(std::cerr);
-      return EXIT_FAILURE;
+    usbguard::IPCClient ipc(/*connected=*/true);
+
+    if (argc == 1) {
+      id = std::stoul(argv[0]);
+      ipc.applyDevicePolicy(id, Rule::Target::Allow, permanent);
+    }
+    else {
```
dkopecek: Does this work with `argc==0`?

src/CLI/usbguard-block-device.cpp
```diff
@@ -75,14 +76,47 @@ namespace usbguard
...
+      for (int i = 1; i < argc; i++) {
+        rule_string.append(" ").append(argv[i]);
+      }
+
+      usbguard::Rule rule;
+      try {
+        rule = Rule::fromString(rule_string);
+      }
+      catch (...) {
+        showHelp(std::cerr);
```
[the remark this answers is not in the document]
Cropi: @radosroka  Is it enough to mention that a RULE_EXCEPTION happened or should it be more specific?

src/CLI/usbguard-allow-device.cpp
Cropi: Yes, you are right. It is a lot more elegant solution.

src/CLI/usbguard-allow-device.cpp
```diff
@@ -75,14 +76,47 @@ namespace usbguard
...
+        return EXIT_FAILURE;
+      }
+
+      /*
+       * Create a ruleset which will contain the requested rule
+      */
+      auto my_rule_set = std::dynamic_pointer_cast<RuleSet>(std::make_shared<MemoryRuleSet>(nullptr));
+      my_rule_set->appendRule(rule);
+
+      for (auto rule_device : ipc.listDevices("match")) {
```
Cropi: I am not sure if the **allow|block|reject-device** commands should affect all the devices or just a few of them, e.g. those which are "allowed", "blocked".
My question is whether I should replace
`for (auto rule_device : ipc.listDevices("match")) {`
with
`for (auto rule_device : ipc.listDevices(argv[0])) {`
@dkopecek

src/CLI/usbguard-block-device.cpp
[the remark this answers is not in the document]
radosroka: It is OK now.

src/CLI/usbguard-allow-device.cpp
```diff
@@ -75,14 +76,47 @@ namespace usbguard
...
+      for (int i = 1; i < argc; i++) {
+        rule_string.append(" ").append(argv[i]);
+      }
+
+      usbguard::Rule rule;
+      try {
+        rule = Rule::fromString(rule_string);
+      }
+      catch (...) {
+        showHelp(std::cerr);
```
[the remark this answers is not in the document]
radosroka: Done.

src/CLI/usbguard-allow-device.cpp
dkopecek: I'm not sure what listDevices(argv[0]) means but I think you can use the match rule constructed from the command line argument directly as a query in listDevices(). But listing all devices and applying the rule only on the matching ones is equivalent. The difference is only which side (cli or daemon) is doing the matching.

Cropi: Sorry, my question was not well formed. 
I choose the first option - the CLI is doing the matching.

dkopecek: Ok

Cropi:
All the requested changes are finished now and is ready for merge.

.travis.yml
```diff
@@ -73,7 +73,7 @@ before_install:
 - sudo apt-get install -y asciidoc
 - sudo apt-get install -y umockdev libumockdev-dev
 - sudo gem install coveralls-lcov
-- curl -JLo /tmp/astyle.tgz --remote-header-name 'https://netix.dl.sourceforge.net/project/astyle/astyle/astyle%203.1/astyle_3.1_linux.tar.gz'
+- curl -JLo /tmp/astyle.tgz --remote-header-name 'netcologne.dl.sourceforge.net/project/astyle/astyle/astyle%203.1/astyle_3.1_linux.tar.gz'
```
radosroka: forgot https://?

tammar96:
Can you please enhance the documentation? An example would be great.

src/Common/Utility.hpp
```diff
@@ -294,6 +294,24 @@ namespace usbguard
...
   };
 
+
+  template<typename T>
+  std::string joinElements(T begin, T end, const std::string& separator = " ")
+  {
+    std::ostringstream ss;
+
+    if (begin != end) {
+      ss << *begin++;
```
radosroka: I think that *(begin++) is much  more readable.

Cropi: Okay

Cropi:
> I tried to test this pull request but I was not able to comprehend what do you want to achieve by the documentation. Can you please enhance the documentation with an usage example?

Some examples related to allow-device have been added.

tammar96:
I tried it and it looks like that it is working flawlessly. :+1:

src/Common/Utility.hpp
tammar96: I agree with @radosroka

radosroka:
Great Thanks.

PR #319, opened 2019-07-23
Move D-Bus support from bridge into usbguard-daemon.

Allen-Webb:


Allen-Webb:
I can take care of the rebase.

Allen-Webb:
I ended up having to move all the glib initialization out of the constructor into the D-Bus IPC server thread.

Allen-Webb:
The change as is doesn't currently provide a way to offer the D-Bus interface on the session bus, it is system only. I don't believe that will be a problem, but I thought I would highlight it in case you all are aware of particular reasons to also support binding to session buses.

src/Library/DBusIPCServerPrivate.cpp
```diff
@@ -339,6 +527,30 @@ namespace usbguard
       with_interface_string.c_str());
     return builder;
   }
+
+  void DBusIPCServerPrivate::addAllowedUID(uid_t uid, const IPCServer::AccessControl& ac)
```
tweksteen: What is the current behavior without these functions implemented? If it is expected for them to be empty, could you add a comment to make that clear? Thanks

src/Library/public/usbguard/IPCServer.cpp
```diff
@@ -281,41 +288,55 @@ namespace usbguard
...
+    for (const auto& d_pointer : d_pointers) {
+      d_pointer->ExceptionMessage(context, object, reason);
+    }
   }
 
   void IPCServer::addAllowedUID(uid_t uid, const IPCServer::AccessControl& ac)
   {
-    d_pointer->addAllowedUID(uid, ac);
+    for (const auto& d_pointer : d_pointers) {
+      d_pointer->addAllowedUID(uid, ac);
```
tweksteen: What's the intended behavior if any of the IPC implementation fail? Should we check the return values?

Allen-Webb: There aren't return values to check. Though I did go through and convert some error log statements to exceptions, so the D-Bus IPC server better matches the behavior of the libqb IPC server code.

src/Library/DBusIPCServerPrivate.cpp
Allen-Webb: I added a comment.

PR #323, opened 2019-08-06
Remove daemon quit method

userWayneCampbell:
In my opinion this function should either be removed or used. I chose remove.


radosroka:
Seems like travis had some network issues... I've restarted CI run.

radosroka:
If anybody will ever need it, he will add it. Thanks!

PR #333, opened 2019-09-24
Add documentation on release process

tweksteen:


radosroka:
I think that the part of this process should be also creation of new release post to [usbguard.github.io](https://usbguard.github.io) via https://github.com/USBGuard/usbguard.github.io repo.

Issue #224, opened 2018-02-25
Request to implement multiple rules.conf files


fcpp31:
Hello,
It would be nice to manage several rules.conf files (one specific for USB keys).
Is it possible to reload the rules.conf file(s) without restarting the usbguard service?
Regards

tammar96:
Hello,
I've got proposal:
- we create a directory /etc/usbguard/rules.d
- there will be multiple rulefiles
- only one implicit rule
- implicit rule has to be defined in usbguard.conf not in rules.conf / rules.d 
- filenames in rules.d have to start with number [0-9][0-9].+ -> that sets the order of the file loading
- rules.conf will be first loaded and then every other file in rules.d by the order mentioned above
- enhance policy to contain multiple rule objects (each one represents one file)
- for interactive allowance there must be one ruleset only for added rules during the runtime

```
rules = [ruleset1, ruleset2, ruleset3, ...]
allRules = union(rules) # concatenate all rules to one
(found, rule) = search(allRules, device): # look for the rule (found is boolean, rule is rule)
if found:
    apply(rule, device)
else
    apply(implicit_rule, device)
```

dkopecek:
@tammar96 :+1: , I would also add:
-  usbguard append-rule & related commands modify only `/etc/usbguard/rules.conf` (or whatever is set via usbguard-dademon.conf). i.e. the drop-in files are not modifiable via the CLI/D-Bus/IPC interfaces.

radosroka:
What about using rules.conf OR content of rules.d directory. If rules.conf is not present and rules.d is empty, daemon will behave the same way like now without rules.conf -> runtime ruleset/policy.

dkopecek:
@radosroka I don't understand exactly what use case you are describing. If rules.conf is not set, then (permanent) modifications via CLI/IPC should fail.

tweksteen:
There is an order when processing the rules. Until now, it has been that the first rule matching is used. I believe we should strictly respect this design when implementing multiple rule files. Having a numeral prefix definitely helps, but we should keep the order in mind during the implementation (your search operation should be deterministic).

For deciding to which file a runtime addition should be targetting, we could use an extra parameter for the name of the rule (and default to rules.conf): `usbguard append-rule -f 01-local.conf "allow"`

tammar96:
I am bit confused right now. 

I wrote:
> for interactive allowance there must be one ruleset only for added rules during the runtime

@dkopecek wrote:
[...]

PR #336, opened 2019-10-22
Implementation of rules.d

tammar96:
Reference: #224

### DIFF
--- a/src/CLI/PolicyGenerator.cpp
+++ b/src/CLI/PolicyGenerator.cpp
@@ -28,7 +28,7 @@ namespace usbguard
 
   PolicyGenerator::PolicyGenerator()
   {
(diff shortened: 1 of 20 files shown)


tammar96:
Please do not merge. Still in progress.

tammar96:
There is a small problem in this case:
```
rules only in rules.d/
[usbguard-daemon running]
usbg append-rule 'allow label "my_label"'
cat rules.conf
[there is every rule from rules.d/ files]

**Comment by tammar96** [comment]
Everything looks good. However when you start the daemon and you want to add the rule via CLI, the rule gets ID which is not right after the last one, e.g.,

```
usbguard list-rules
1. ...
2. ...
3. ...
```

```
usbguard append-rule ... -a 3
```
or without parameter `-a`
```
usbguard append-rule ... 
```

```
usbguard list-rules
1. ...
2. ...
3. ...
7. ...
```

I am unable to find out why this behavior is happening. Could you help me to find out what's wrong?

PR #339, opened 2019-10-28
Replace /var with localstatedir placeholder in systemd unit file

tu-maurice:
In certain environments one cannot assume the localstatedir to be /var. Gentoo e.g. uses /var/lib. I don't know if this is even the right fix or if localstatedir might be misused for files that are supposed to got into /var/log.


tu-maurice:
^ forgot to replace localstatedir in usbguard.service

Makefile.am
```diff
@@ -115,7 +115,11 @@ CLEANFILES+=\
 	$(top_builddir)/usbguard.service
 
 usbguard.service: $(top_srcdir)/usbguard.service.in
-	$(SED) -e "s|%sbindir%|${sbindir}|" -e "s|%sysconfdir%|${sysconfdir}|" $^ > $@ || rm -f $@
+	$(SED) \
+	  -e "s|%sbindir%|${sbindir}|" \
+	  -e "s|%sysconfdir%|${sysconfdir}|" \
+	  -e 's|%localstatedir%|${localstatedir}|" \
```
radosroka: Could you replace the first ```'``` wtih ```"``` otherwise it is not valid.

tu-maurice: Oh, sure. My bad.

radosroka:
Thanks.

PR #342, opened 2019-11-13
Add match-all keyword

tammar96:
Keyword match-all extends functionality of the current grammar.
Example:
```allow id *:* with-interface all-of { 03:*:* 09:*:* }```
this will check if the interfaces from set `{ 03:*:* 09:*:* }` are present in the device in other words if the set `{ 03:*:* 09:*:* }` is subset of the device interfaces.

In the other hand rule like this: ```allow id *:* with-interface match-all { 03:*:* 09:*:* }``` will check if all of the interfaces in the device are subset of the set { 03:*:* 09:*:* }.

The difference is that in the case of the `match-all` keyword we can have device like keyboard `{03:00:01}` and match it against that one rule (```allow id *:* with-interface match-all { 03:*:* 09:*:* }```).

### DIFF
--- a/doc/man/usbguard-rules.conf.5.adoc
+++ b/doc/man/usbguard-rules.conf.5.adoc
@@ -93,6 +93,9 @@ where the optional 'operator' is one of:
 *equals-ordered*::
     The device attribute set must contain exactly the same set of values in the same order for the rule to match.
 
+*match-all*::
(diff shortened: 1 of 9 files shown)


tammar96:
@tweksteen @radosroka Can you please check if it works properly? I tried it, luckily I haven't encounter any problems but let's be sure.

PR #344, opened 2019-11-19
Update travis to newer compilers

tammar96:


tammar96:
Is there any way to have travis check in this PR?

PR #351, opened 2020-01-05
configure.ac: Fix libcap-ng not found message

hartwork:
For your consideration :smile:


radosroka:
Thanks for your contribution :)

PR #361, opened 2020-02-11
Public API documentation

### LINKED ISSUE #327: Missing API documentation
Usbguard API is frustruating and contra-productive to use due to the fact that there is little or no documentation for it and API user has to search in the sources of usbguard to find out what each function is doing. This is a huge issue of usability of the usbguard API.

ZoltanFridrich:
Closes #327

### DIFF
--- a/src/Library/public/usbguard/KeyValueParser.cpp
+++ b/src/Library/public/usbguard/KeyValueParser.cpp
@@ -35,10 +35,6 @@
 
 namespace usbguard
 {
-  /**
-  * NOTE:
-  * vector v should have strings UPPERCASE when case insensitive mode was enabled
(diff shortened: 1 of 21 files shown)


src/Library/public/usbguard/KeyValueParser.cpp
```diff
@@ -35,10 +35,6 @@
 
 namespace usbguard
 {
-  /**
```
tammar96: Why did you remove this note?

ZoltanFridrich: Because its included in the hpp file documentation and it was only polluting doxygen generated documentation.

PR #368, opened 2020-03-20
Add configure flags for optional libraries.

Allen-Webb:
This enables distributions to force disable features based on optional dependencies to avoid unexpected behavior when extra headers are present.


radosroka:
Thanks!

PR #371, opened 2020-04-05
Drop dbus-glib dependency

bisco2:
dbus-glib is deprecated and the configure script doesn't seem to use the library, but sometimes incorrectly refers to it although using GDBus

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -325,17 +325,17 @@ AC_LANG_POP
 CPPFLAGS=$SAVE_CPPFLAGS
 
 #
-# GLib D-Bus
+# GDBus
 #
 AC_ARG_WITH([dbus], AC_HELP_STRING([--with-dbus], [Build the DBus Bridge service]), [], [with_dbus=yes])
 if test "x$with_dbus" = xyes; then
   #
   # Check for required D-Bus modules
   #
-  PKG_CHECK_MODULES([dbus], [dbus-1 dbus-glib-1 >= 0.100 gio-2.0],
-  [AC_DEFINE([HAVE_DBUS], [1], [Required GLib DBus API available])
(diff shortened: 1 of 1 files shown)


smcv:
If you aren't using `<dbus/dbus.h>` then you also don't need to check for the pkg-config module `dbus-1`. To use the GDBusConnection, GDBusProxy, etc. APIs from `<gio/gio.h>`, it's sufficient to check for the pkg-config module `gio-2.0`.

bisco2:
> If you aren't using `<dbus/dbus.h>` then you also don't need to check for the pkg-config module `dbus-1`. To use the GDBusConnection, GDBusProxy, etc. APIs from `<gio/gio.h>`, it's sufficient to check for the pkg-config module `gio-2.0`.

I think the `dbus-1` dependency is there to be able to find out the `dbus_services_dir` (`system_bus_services_dir`) in:
https://github.com/USBGuard/usbguard/blob/29bee4b9f3efdba1960d1a95f83258ab0989edd4/configure.ac#L363 and the `dbus_data_dir` in:
https://github.com/USBGuard/usbguard/blob/29bee4b9f3efdba1960d1a95f83258ab0989edd4/configure.ac#L386

smcv:
> I think the dbus-1 dependency is there to be able to find out the dbus_services_dir 

Fair enough. I would personally hard-code `dbus_services_dir` to `${datadir}/dbus-1/system-services` (which is where https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-types-system says they should go, so it's a long-term-stable path); but if you want to query it from a package, then yes you'll need `dbus-1`.

radosroka:
Thanks!

PR #377, opened 2020-04-27
Drop unused PIDFile statement from service file

bisco2:
The service file does not start usbguard in daemon mode so the PIDFile is never creatd. Therefore the PIDFile statement in the service file is not needed and might lead to confusion.


radosroka:
Yes, you are right. Thanks!

Issue #353, opened 2020-01-09
PresenceChanged signal does not include final policy target (i.e. block or allow)


muelli:
I have a simple setup:

```
$ sudo cat /etc/usbguard/usbguard-daemon.conf | grep -v '#' | tr -s '\n'
RuleFile=/etc/usbguard/rules.conf
ImplicitPolicyTarget=allow
PresentDevicePolicy=allow
PresentControllerPolicy=keep
InsertedDevicePolicy=apply-policy
RestoreControllerDeviceState=false
DeviceManagerBackend=uevent
IPCAllowedUsers=root muelli
IPCAllowedGroups=root
IPCAccessControlFiles=/etc/usbguard/IPCAccessControl.d/
DeviceRulesWithPort=false
AuditFilePath=/var/log/usbguard/usbguard-audit.log
$ usbguard  list-rules | wc -l
0
$
```

This is what usbguard watch has to say when I re-plug a device:

```
[IPC] Connected
[device] PresenceChanged: id=73
 event=Remove
 target=allow
 device_rule=allow id 0b0e:0422 serial "1C48F97F3BA7020A00" name "Jabra SPEAK 510 USB" hash "zibEDXpRn6PH9/nNbOsNvJ1+S9QInKkQvTQO5kZz0Eo=" parent-hash "jEP/6WzviqdJ5VSeTUY8PatCNBKeaREvo2OqdplND/o=" via-port "1-1" with-interface { 01:01:00 01:02:00 01:02:00 01:02:00 01:02:00 03:00:00 }
[device] PresenceChanged: id=74
 event=Insert
 target=block
 device_rule=block id 0b0e:0422 serial "1C48F97F3BA7020A00" name "Jabra SPEAK 510 USB" hash "zibEDXpRn6PH9/nNbOsNvJ1+S9QInKkQvTQO5kZz0Eo=" parent-hash "jEP/6WzviqdJ5VSeTUY8PatCNBKeaREvo2OqdplND/o=" via-port "1-1" with-interface { 01:01:00 01:02:00 01:02:00 01:02:00 01:02:00 03:00:00 }
[device] PolicyChanged: id=74
 target_old=block
 target_new=allow
[...]

PR #385, opened 2020-05-28
Always throw the DevicePolicyChanged signal

muelli:
The motivation is to have a simple "this is what happens with the device" signal, e.g. whether a device is accepted or rejected. This makes writing front-ends easier. I dare to say that this is the precondition to make it possible at all. As mentioned in https://github.com/USBGuard/usbguard/issues/353 it is difficult, at best, to write something useful with the current behaviour of the signal, which is a bit hard to explain.
Right now, the signal is being thrown when a device is present, inserted, or updated, except when a rule has matched that has changed whatever the implicit rule said it would do.  If I want to know whether an inserted device is actually enabled, I need to listen to the PresenceChanged signal and then wait for the PolicyChanged signal. Except that it doesn't necessarily come which makes it hard to follow.

I have tried to find consumers of this signal and I have [found three](https://codesearch.debian.net/search?q=DevicePolicyChanged&literal=1). Chrome OS, GNOME, and [py3status](https://sources.debian.org/src/py3status/3.22-1/py3status/modules/usbguard.py/?hl=98#L98). I don't know whether and how Chrome OS will suffer from this change, but it could be that it breaks something on their end. It will certainly help GNOME. And after having looked at what py3status wants to achieve, my guess is that it will benefit from this change, too. From what I can see, they don't want the current behaviour but, much like GNOME, something that tells them what happens to an inserted device.

I appreciate that changing the behaviour of a signal is frowned upon. But I question the usefulness of the current signal and I doubt that Chrome OS has a usecase that requires this signal (but again: I haven't checked).
Introducing a new signal is certainly possible and would probably be the cleanest approach. But I know little about USBGuard and just moving the emission of the signal out of the guard is a really simple change in the code. A new signal seems to involve a lot of boilerplate for the multi-layer IPC abstraction. No fun.

Another option is to rename the signal, i.e. to remove the existing signal and introduce a new one. This *should* be fairly easy with `sed`. But it will at least actively break the existing consumers. And it might break more things that we don't know of yet. Not that this is a good reason to not do the correct thing, but I thought I'd come up with the easiest solution for me and listen to what you think.

### DIFF
--- a/src/DBus/DBusInterface.xml
+++ b/src/DBus/DBusInterface.xml
@@ -213,7 +213,11 @@
                      ImplicitPolicyTarget or InsertedDevicePolicy.
        @attributes: A dictionary of device attributes and their values.
 
-      Notify about a change of a USB device authorization target.
+      This signal fires whenever a device is present (i.e. during startup
+      of the daemon), inserted (i.e. plugged in), or updated (i.e. when a rule
+      has been updated).
+
+      This signal is poorly named. It should probably be called "RuleApplied".
 
       The device attribute dictionary contains the following attributes:
         - id (the USB device ID in the form VID:PID)
--- a/src/Daemon/Daemon.cpp
+++ b/src/Daemon/Daemon.cpp
@@ -854,16 +854,16 @@ namespace usbguard
       else {
         USBGUARD_LOG(Debug) << "Implicit rule matched";
       }
-
-      std::shared_ptr<const Rule> device_rule = \
-        device_post->getDeviceRule(/*with_port=*/true,
-          /*with_parent_hash=*/true);
-      DevicePolicyChanged(device->getID(),
-        target_old,
-        device_post->getTarget(),
-        device_rule->toString(),
-        matched_rule->getRuleID());
     }
+
+    std::shared_ptr<const Rule> device_rule = \
+      device_post->getDeviceRule(/*with_port=*/true,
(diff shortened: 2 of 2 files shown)


muelli:
@radosroka any opinion on the change? Or rather the goal of the change (as I appreciate that it may not be the best solution for solving the problem)?

radosroka:
Hi @muelli,

I can see your point here and I would like to do something about that. 
But still, we need to somehow preserve backwards compatibility which is not easy task in this case.

I'll be scrolling the code little bit and I let you know when I come up with something.

muelli:
Hi @radosroka 
thanks for your response. And I appreciate that backwards compatibility is a concern. It's not clear to me yet, though, whether you are okay with loudly breaking app by, e.g. renaming the signal or bumping the version number. If you want to maintain behavioural compatibility then the only option I see is to introduce a new signal. That is fine with me. I'm willing to cook up a patch. I just want to test waters before investing the work.

radosroka:
@pinotree what do you think as a new applet developer?

pinotree:
> @pinotree what do you think as a new applet developer?

Sorry, I don't have that much of experience yet with the usbguard code. Can you please explain what are the implicit rules?

muelli:
this has been superseded by https://github.com/USBGuard/usbguard/pull/459

PR #390, opened 2020-06-24
Use forking style when running with systemd

radosroka:
- with such configuration "systemctl start usbguard" can return error code
  in case of bad configuration

Signed-off-by: Radovan Sroka <rsroka@redhat.com>


radosroka:
@tweksteen are you familiar with any other side effect?

PR #391, opened 2020-07-06
doc: include defaults in the description

muelli:
The intention of this change is to make it easier to understand what
USBGuard does by looking at the docs, in particular the Web site.
The defaults have been taken from usbguard-daemon.conf.in. I haven't
noticed any automatic way to generated the (part) of the documentation
from that file, although it might be clever to do in order to lower the
redundancy.

Note that I haven't built the docs, mainly, because I don't know how to, and thus I haven't checked whether it looks nice enough and whether it actually lands in the generated Web sites served at https://usbguard.github.io/documentation/configuration.html#options

### DIFF
--- a/doc/man/usbguard-daemon.conf.5.adoc
+++ b/doc/man/usbguard-daemon.conf.5.adoc
@@ -18,6 +18,7 @@ It may be overridden using the *-c* command-line option, see *usbguard-daemon*(8
 *RuleFile*='path'::
     The USBGuard daemon will use this file to load the policy rule set from it
     and to write new rules received via the IPC interface.
+    Default: %sysconfdir%/usbguard/rules.conf
 
 *RuleFolder*='path'::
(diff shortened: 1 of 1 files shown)


radosroka:
> The intention of this change is to make it easier to understand what
> USBGuard does by looking at the docs, in particular the Web site.
> The defaults have been taken from usbguard-daemon.conf.in. I haven't
> noticed any automatic way to generated the (part) of the documentation
> from that file, although it might be clever to do in order to lower the
> redundancy.
> 
> Note that I haven't built the docs, mainly, because I don't know how to, and thus I haven't checked whether it looks nice enough and whether it actually lands in the generated Web sites served at https://usbguard.github.io/documentation/configuration.html#options

There is separate repo for the website: https://github.com/USBGuard/usbguard.github.io

radosroka:
LGTM.

PR #411, opened 2020-08-31
Fix oss-fuzz build, by adding unistd.h to FDInputStream.hpp.

Allen-Webb:
This resolves the following build error for usbguard on oss-fuzz:

Step #4: In file included from src/Library/DeviceBase.cpp:26:
Step #4: ./src/Common/FDInputStream.hpp:65:15: error: use of undeclared identifier 'read'
Step #4:         ret = read(fd_, s, n);


radosroka:
Thanks!

PR #425, opened 2020-10-19
Remove root permission check option

ZoltanFridrich:
No root check option is currently not working due to the fact that CLI will still need root permissions to read usbguard-daemon.conf file. This option should be removed as there does not seem to be any rational reason for its existence within usbguard.


radosroka:
I definitely agree with you. I'll merge it.

PR #426, opened 2020-10-21
Extend public API - Allow IPC Client to query IPC permissions

Cropi:
Allow IPC Client to ask IPC server for a certain AccessControl(Section, Privilege). To get a better understanding about this problem I implemented the ```check-IPC-permission``` CLI command to test the functionality.
Syntax: ```usbguard check-IPC-permission --section Policy --privilege modify```
This can be deleted after change is aproved and before it is merged.

This would help with https://github.com/Cropi/usbguard-notifier/pull/62#issuecomment-681157855

Help needed:
- [x] Right now, it is not possible to check IPC permission for ALL and NONE section and privilege. **Maybe** we should extend ```IPCServer::AccessControl::hasPrivilege```  with that. The reason I did not do that is it might have some unexpected consequences.
- [x] The use of ```checkIPCpermission(section, privilege)``` requires Section=ALL, Privilege=None permissions, is this the right approach?

### DIFF
--- a/src/Library/IPCServerPrivate.cpp
+++ b/src/Library/IPCServerPrivate.cpp
@@ -84,6 +84,8 @@ namespace usbguard
       IPCServer::AccessControl::Privilege::MODIFY);
     registerHandler<IPC::getParameter>(&IPCServerPrivate::handleGetParameter, IPCServer::AccessControl::Section::PARAMETERS,
       IPCServer::AccessControl::Privilege::LIST);
+    registerHandler<IPC::checkIPCPermissions>(&IPCServerPrivate::handleCheckIPCPermissions, IPCServer::AccessControl::Section::ALL,
+      IPCServer::AccessControl::Privilege::NONE);
   }
 
   void IPCServerPrivate::initIPC()
@@ -987,6 +989,22 @@ namespace usbguard
     IPC::getParameter* const message_out = message_in->New();
     message_out->MergeFrom(*message_in);
     message_out->mutable_response()->set_value(value);
+    response.reset(message_out);
+  }
+
+  void IPCServerPrivate::handleCheckIPCPermissions(IPC::MessagePointer& request, IPC::MessagePointer& response)
+  {
+    const IPC::checkIPCPermissions* const message_in = static_cast<const IPC::checkIPCPermissions*>(request.get());
+    uid_t uid = message_in->request().uid();
+    gid_t gid = message_in->request().gid();
+    IPCServer::AccessControl access_control = IPCServer::AccessControl();
+    const bool auth = qbIPCConnectionAllowed(uid, gid, &access_control);
+    IPCServer::AccessControl::Section section = IPCServer::AccessControl::sectionFromString(message_in->request().section());
+    IPCServer::AccessControl::Privilege privilege = IPCServer::AccessControl::privilegeFromString(message_in->request().privilege());
+    const bool permit = auth && access_control.hasPrivilege(section, privilege);
+    IPC::checkIPCPermissions* const message_out = message_in->New();
+    message_out->MergeFrom(*message_in);
+    message_out->mutable_response()->set_permit(permit);
     response.reset(message_out);
   }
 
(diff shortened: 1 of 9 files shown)


src/Library/IPCServerPrivate.cpp
```diff
@@ -84,6 +84,8 @@ namespace usbguard
       IPCServer::AccessControl::Privilege::MODIFY);
     registerHandler<IPC::getParameter>(&IPCServerPrivate::handleGetParameter, IPCServer::AccessControl::Section::PARAMETERS,
       IPCServer::AccessControl::Privilege::LIST);
+    registerHandler<IPC::checkIPCPermissions>(&IPCServerPrivate::handleCheckIPCPermissions, IPCServer::AccessControl::Section::ALL,
+      IPCServer::AccessControl::Privilege::NONE);
```
Cropi, line 88: ``IPCServer::AccessControl::Section::ALL, IPCServer::AccessControl::Privilege::NONE``
Is this the right approach?

src/CLI/usbguard-check-IPC-permission.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+      default:
+        return EXIT_FAILURE;
+      }
+    }
+
+    try {
+      IPCServer::AccessControl::Section section = IPCServer::AccessControl::sectionFromString(section_string);;
+      IPCServer::AccessControl::Privilege privilege = IPCServer::AccessControl::privilegeFromString(privilege_string);;
+      usbguard::IPCClient ipc(/*connected=*/true);
```
ZoltanFridrich, line 83: Afaik IPCClient needs to be able to connect to IPC interface when its initialized. Is it really a good approach that you need IPC connection to check IPC permission? Maybe it would be better to have permissions for each operation stored outside IPCServer so the user would not require IPC connection to check permissions.

src/Library/IPCServerPrivate.cpp
ZoltanFridrich: Makes sense imo.

src/CLI/usbguard-check-IPC-permission.cpp
Cropi: The idea sounds good to me but there is a problem with that: only IPCServer knows the IPC permissions, so in order to get them it is inevitable to use IPC connection. One possible way to do that without using the IPCClient would be to read it from configuration files but it would require root privileges, so I think it's not an option to consider.

Cropi:
Hi @LucasParsy,

This could help with https://github.com/Cropi/usbguard-notifier/pull/62. You can check if user has ```Policy=modify``` IPC Access Control privilege via```ipc.checkIPCpermission(section, privilege)``` method.
What do you think? Is it enough?

doc/man/usbguard.1.adoc
```diff
@@ -39,6 +39,8 @@ usbguard add-user 'name'
 
 usbguard remove-user 'name'
 
+usbguard check-IPC-permission
```
ZoltanFridrich, line 42: Why are you making this a part of usbguard CLI? I thought this is just a way for developers to check IPC permissions before trying and possibly failing a call to daemon. In my opinion this should not be a part of CLI.

radosroka (approved):

LucasParsy:
Hey @Cropi ! Answering a bit late, but thanks for your work, it would indeed really help.

I tried using it in my code but unfortunately it doesnt work.
It fails silently, so I don't know where it could or I could have gone wrong.

i put a call to `this->checkIPCPermissions(IPCServer::AccessControl::Section::POLICY, IPCServer::AccessControl::Privilege::MODIFY)` 

It does compile, but code after this on my method is not executed.
It doesn't trigger any error message on the usbguard daemon or the notifier.

I am using the latest version of usbguard, I made a `sudo make install`
I tried running usbguard-notifier with `LD_PRELOAD="{path to my usbgard cloned git directory}/.libs/libusbguard.so"`
I tried running usbguard-notifier (and ubsguard daemon) as sudo user.

any idea of where could the problem be?

Cropi:
Hey @LucasParsy,

I provided a possible solution at https://github.com/Cropi/usbguard-notifier/pull/62. Let's continue the discussion there :)

PR #440, opened 2020-12-08
Do not call processDevicePresence when CHANGE uevent is invoked

Cropi:
Commit https://github.com/USBGuard/usbguard/commit/f87728ddc679ab5f14a4572adde5f97c760732fc : in some cases, when the policy of a device is changed(e.g. allow -> block), a CHANGE uevent action is invoked. Once it happens, usbguard calls ```processDevicePresence()``` and it changes back the policy(e.g. block -> allow) of that specific device. This is still in progress as I am not not sure about this.
I did some experiments with ```udevadm monitor``` and draw conclusions that we might want to remove action=CHANGE events.

### DIFF
--- a/src/Library/UEventDeviceManager.cpp
+++ b/src/Library/UEventDeviceManager.cpp
@@ -345,7 +345,7 @@ namespace usbguard
       uint32_t id = 0;
       const bool known_path = knownSysfsPath(sysfs_devpath, &id);
 
-      if (action == "add" || action == "change") {
+      if (action == "add" /*|| action == "change"*/) {
         if (known_path && id > 0) {
           processDevicePresence(id);
         }


ZoltanFridrich:
I do confirm that this change fixes the bug where for some devices if you want to block a specific device, nothing happens. What happens is that the device gets correctly blocked on the system. This generates uevents with ACTION=remove which gets processed and the device is then removed from the device map. However, for some devices not only ACTION=remove uevents are generated but also ACTION=change. Uevents with ACTION=change are processed and usbguard tries to apply policy for such devices. Therefore the device which has been just blocked will get immediately allowed if this device is allowed in the usbguard policy. I tried multiple devices and the results are consistent. However ACTION=change uevents are generated only for some devices, so some devices can be blocked even if they are allowed in the policy and some can't. I also tested older usbguard versions where I know for a fact that I didn't have this issue and these versions suffer from this too, so this was not caused by any usbguard change.

tl;dr:
Not applying policy for ACTION=change uevents will fix the issue. However, I am not sure what is the security impact.

radosroka:
> I do confirm that this change fixes the bug where for some devices if you want to block a specific device, nothing happens. What happens is that the device gets correctly blocked on the system. This generates uevents with ACTION=remove which gets processed and the device is then removed from the device map. However, for some devices not only ACTION=remove uevents are generated but also ACTION=change. Uevents with ACTION=change are processed and usbguard tries to apply policy for such devices. Therefore the device which has been just blocked will get immediately allowed if this device is allowed in the usbguard policy. I tried multiple devices and the results are consistent. However ACTION=change uevents are generated only for some devices, so some devices can be blocked even if they are allowed in the policy and some can't. I also tested older usbguard versions where I know for a fact that I didn't have this issue and these versions suffer from this too, so this was not caused by any usbguard change.
> 
> tl;dr:
> Not applying policy for ACTION=change uevents will fix the issue. However, I am not sure what is the security impact.

From my understanding of the code, it should be OK. We should care only for `add` and `remove` actions.

radosroka (approved):

PR #454, opened 2021-02-18
Increase C++ version to c++17

ZoltanFridrich:
Increasing c++ version from 11 to 17 so that we are able to use newer versions of development packages, namely PEGTL 3+.
After investigation, there does not seem to be any compatibility issues.


Cropi:
I've compiled usbguard from sources with c++17 and I did not encounter any problems. Looks good to me.

Cropi (approved):

PR #458, opened 2021-03-03
Apply PEGTL 3.2.0 fixes to fuzzers.

Allen-Webb:
This fixes the ossfuzz build for usbguard.


Allen-Webb:
As a side note, dkopecek@ is the only usbguard contact for ossfuzz alerts. It might make sense to get other maintainers added to the list.

radosroka:
> As a side note, dkopecek@ is the only usbguard contact for ossfuzz alerts. It might make sense to get other maintainers added to the list.

Since @dkopecek  does not work on the usbguard anymore, we should change that.

PR #459, opened 2021-03-08
Add new DevicePolicyApplied signal

muelli:
The intention is to fix https://github.com/USBGuard/usbguard/issues/353, e.g. have a reliable signal to know what has happened to a device. This is the hard way of what I described [here](https://github.com/USBGuard/usbguard/pull/385#issue-424670552), i.e. a new signal rather than changing existing behaviour.

If this is acceptable, I may be working on docs. But since it's unclear to me that this is an acceptable change, I haven't taken the time for writing docs, yet.

For now, this is WIP, because of the commits enabling the logging. Also, I had a silly mistake when defining the new signal. Fortunately, @RyuzakiKK beat me to it. I will, if this is an acceptable change, separate the logging and squash commits.

@radosroka WDYT?

### DIFF
--- a/src/DBus/DBusInterface.xml
+++ b/src/DBus/DBusInterface.xml
@@ -152,8 +152,8 @@
       If the permanent flag is set to True, a rule will be appended to the policy or an exiting device
       rule will be modified in order to permanently store the authorization decision.
 
-      Sucessfull exection of this method will cause the DevicePolicyChanged signal to be broadcasted if
-      the device authorization target was different than the applied target.
+      Successful execution of this method will cause the DevicePolicyChanged signal to be broadcasted if
+      the device authorization target was different from the applied target.
       -->
     <method name="applyDevicePolicy">
       <arg name="id" direction="in" type="u"/>
@@ -205,6 +205,7 @@
       DevicePolicyChanged:
        @id: Device id of the device
        @target_old: Previous authorization target in numerical form.
+                    0 = Allow. 1 = Block. 2 = Reject.
        @target_new: Current authorization target in numerical form.
        @device_rule: Device specific rule.
        @rule_id: A rule id of the matched rule. Otherwise a reserved rule id value is used.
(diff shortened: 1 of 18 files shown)


muelli:
@Cropi do you have an opinion on this patch?

Cropi:
Hi @muelli,

This sounds like a good idea to me. Maybe it's the hard way of solving this issue but at least backwards compatibility will be ensured(really appreciate it). If by any chance you think there are too many signals, we can somehow(in the future):

- Extend ```usbguard watch```, so it can ignore some of the signals.
- Extend the daemon configuration, so it emits only specified signals.

muelli:
Thanks for your response :)

I don't think the number of signals is too high, yet.
But I do think that consumers are tempted to believe the wrong things about them, as I discussed in the other bug.

Let me know what needs to be done in order to get this merged.

Cropi:
> Thanks for your response :)
> 
> I don't think the number of signals is too high, yet.
> But I do think that consumers are tempted to believe the wrong things about them, as I discussed in the other bug.

I agree.

> 
> Let me know what needs to be done in order to get this merged.

Overall, it looks good to me. I will do a few manuals tests with it a little bit later. In case I find something that should be addressed, I will let you know. Otherwise, I will merge it :)
Thanks.

muelli:
@ZoltanFridrich do you have an opinion here? Can I do anything to move this forward?

src/DBus/DBusInterface.xml
```diff
@@ -152,8 +152,8 @@
       If the permanent flag is set to True, a rule will be appended to the policy or an exiting device
       rule will be modified in order to permanently store the authorization decision.
 
-      Sucessfull exection of this method will cause the DevicePolicyChanged signal to be broadcasted if
-      the device authorization target was different than the applied target.
+      Sucessful execution of this method will cause the DevicePolicyChanged signal to be broadcasted if
```
ZoltanFridrich, line 155: successful, not sucessful

ZoltanFridrich:
IMO this change is acceptable already. The change looks good and makes sense. The commits should be squashed and maybe actually fix the minor typo. However, I am not an USBGuard upstream maintainer. It needs to be approved by either @radosroka or @Cropi

ZoltanFridrich (approved):

Cropi (approved):
Thanks again for the PR. Could you squash the commits? I will merge it then.

muelli:
@Cropi I have left separate one commit with additions to the logging. I'm also happy to squash it, if you want me to.

Cropi (approved):

Issue #465, opened 2021-03-22
USBGuard compilation failed with lower OpenSSL versions prior to 1.1.0


anudeepthika:
I have tried installing USBGuard with OpenSSL 1.0.2 as the crypto-backend. The compilation ended with the below message.

> | src/Library/Hash.cpp: In constructor 'usbguard::Hash::Hash()':
> | src/Library/Hash.cpp:42:19: error: 'EVP_MD_CTX_new' was not declared in this scope
> |      if ((_state = EVP_MD_CTX_new()) == nullptr)
> |                         ^~~~~~~~~~~~~~
> | src/Library/Hash.cpp: In destructor 'usbguard::Hash::~Hash()':
> | src/Library/Hash.cpp:104:5: error: 'EVP_MD_CTX_free' was not declared in this scope
> |      EVP_MD_CTX_free(_state);
> |      ^~~~~~~~~~~~~~~

EVP_MD_CTX_new() in 1.1.0 has replaced EVP_MD_CTX_create() in 1.0.x.

EVP_MD_CTX_free() in 1.1.0 has replaced EVP_MD_CTX_destroy() in 1.0.x.

genodeftest:
As the pull request is merged, can you please close this issue?

PR #466, opened 2021-03-22
Add support for lower OpenSSL versions prior to 1.1.0

anudeepthika:
-Commit e4bdd55a (Add OpenSSL as crypto backend for device hashing) introduced
support for newer versions of OpenSSL only
-Fix this by detecting the OpenSSL version during the compile and use the proper
EVP digest routines 

Link to the issue: #465


Cropi (approved):
LGTM. Thanks for the Pull Request.

PR #469, opened 2021-04-08
Add #include <tao/pegtl.hpp> to fuzzers.

Allen-Webb:
tao/pegtl.hpp includes version.hpp which defines
TAO_PEGTL_VERSION_MAJOR. This is needed for #458 to work as intended.


Cropi (approved):
Thanks.

PR #479, opened 2021-06-25
Improve group/user name checking

### LINKED ISSUE #478: Accept user with non alpha characters in add-user
My username is based on my real name, namely "jean-luc". But usbguard add-user jean-luc report the followin error: ERROR: IPC access control: name contains invalid character(s): jean-luc

Please allow to use this "invalid" character.

ZoltanFridrich:
Make name checking consistent with **useradd** manual page definition (with addition of capital letters).

Resolves #478

### DIFF
--- a/src/Common/Utility.cpp
+++ b/src/Common/Utility.cpp
@@ -542,6 +542,32 @@ namespace usbguard
     std::sort(rulefile_list.begin(), rulefile_list.end());
     return rulefile_list;
   }
+
+  bool isValidName(const std::string& name)
+  {
+    const char* s = name.data();
+
+    if (('\0' == *s) ||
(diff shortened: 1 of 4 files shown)


Cropi (approved):
Looks good to me as it is consistent with ```useradd``` command.

JeanLucCoulon:
From the manual page of useradd, it seems to me (at least on debian), that a dash (-) is allowed:
"It is usually recommended to only use usernames that begin with a lower
case letter or an underscore, followed by lower case letters, digits,
underscores, or dashes. They can end with a dollar sign. In regular
expression terms: [a-z_][a-z0-9_-]*[$]?"

J-L

PR #480, opened 2021-06-30
[WIP] Add insertRule method

### LINKED ISSUE #471: Append a rule at the beginning of a ruleset?
append-rule requires a parent id, making inserting a rule at the beginning impossible. Is there any work around or can we have a prepend-rule method? Thanks!

ZoltanFridrich:
Currently, appendRule function allows appending rules either to the end of the last ruleset or after a rule with the given id. This behaviour is not very flexible as you cant append a rule to the beginning of any ruleset and you are unable to append a rule into an empty ruleset. To improve this, rulesets can now be referred to by their filename. And instead of changing the appendRule method a new API call InsertRule has been added.

InsertRule method will receive a rule and optional arguments: rule_id (same as parent_id in appendRule, serves like an anchor for appending), ruleset_prefix (this string is used to match a prefix of a ruleset name, basically helps you to chose into which ruleset the rule should be inserted), insert_before (flag which is used to determine whether a rule should be appended before or after a rule with ID rule_id)

After this addition, two new options can be added to usbguard append-rule command:
--before (-b) id : same as --after option but you are appending before a rule with given id
--ruleset (-s) prefix : prefix of a ruleset where the rule should be inserted

The addition of named rulesets also opens more options for enhancements into the future, for example, usbguard list-rules might be able to separate output based on rulesets or print only rules within a separate ruleset.

Resolves #471


muelli:
I wonder whether easier ways exist to achieve "prepend".
Is allowing `0` as a parent id problematic?

ZoltanFridrich:
> I wonder whether easier ways exist to achieve "prepend".
> Is allowing `0` as a parent id problematic?

Actually, it shouldn't be problematic at all. However, you need to be able to specify the ruleset anyway as you would not be able to prepend to any ruleset other than the first. I will remove --before option and rather make it such that --after 0 means inserting the rule at the beginning.

muelli:
> However, you need to be able to specify the ruleset anyway

hm. I'm a bit lost here.
The DBus interface doesn't require a ruleset, does it? AFAICS, it's `appendRule (IN  s     rule,
            IN  u     parent_id,
            OUT u     id);`.

And I believe I have successfully used "0" to prepend in the past.
Or was I just lucky and have always had empty rules, anyway?

ZoltanFridrich:
> And I believe I have successfully used "0" to prepend in the past.
> Or was I just lucky and have always had empty rules, anyway?

It might have been possible in the past as the underlying code works that way, however with current version it should be impossible as it checks for the presence of rule with parent_id (which will always be missing for 0) and the append-rule fails. If I remember it correctly.

src/DBus/DBusInterface.xml
```diff
@@ -74,6 +74,30 @@
       <arg name="ruleset" direction="out" type="a(us)"/>
     </method>
 
+    <!--
+      insertRule:
+       @rule: The rule that should be appended to the policy.
+       @parent_id: Rule id of the parent rule.
+       @ruleset: Prefix of a ruleset where the rule should be appended.
```
muelli, line 81: I am confused with the concept of rulesets. The only other place where this is exposed seems to be the `org.usbguard.Policy::listRules` call: https://usbguard.github.io/documentation/dbus/doc-org.usbguard.Policy.html#gdbus-method-org-usbguard-Policy.listRules
I can't find the term anywhere else.
As such, I wonder how I as a user would know what string to provide. Is the empty string a valid input? Where do I get the other "prefixes" from to make a choice?

muelli:
> however with current version it should be impossible as it checks for the presence of rule with parent_id (which will always be missing for 0) and the append-rule fails. If I remember it correctly.

yeah, I does seem to fail. USBGuard must have changed behaviour.
But I fail to see where the failure is coming from. From looking at

https://github.com/USBGuard/usbguard/blob/034f3786082bf84563907254ebae05ff50eb103a/src/Library/public/usbguard/RuleSet.cpp#L104-L105

I can see the intention the give "0" the "prepending" semantic. What am I missing?

ZoltanFridrich:
> I can see the intention the give "0" the "prepending" semantic. What am I missing?

It was implemented in RuleSet::appendRule but not in Policy::appendRule from which the RuleSet::appendRule method is called. The Policy::appendRule method checks whether the given id exists within the ruleset, but the rule with id 0 can never exist within a ruleset. Thats why it fails.

muelli:
Thanks for explaining. I think I see how the code rejects parent rule id 0.
This probably fails:
https://github.com/USBGuard/usbguard/blob/034f3786082bf84563907254ebae05ff50eb103a/src/Library/public/usbguard/Policy.cpp#L78


It seems that the [rules.d changes](https://github.com/USBGuard/usbguard/commit/a0daeeb993d7aff77ff02c13bef5094a8860d9a7) broke the behaviour of prepending by appending with parent rule id 0.

I guess it would be easy to restore the behaviour by checking for "0", just as RuleSet::appendRule does.

ZoltanFridrich:
> I guess it would be easy to restore the behaviour by checking for "0", just as RuleSet::appendRule does.

Yes, it would be, but I wanted to expand the usbguard-cli to work with policy files in rules.d folder.

ZoltanFridrich:
I have stopped to work on this PR as I am moving to other projects.

This PR is supposed to enhance users ability to append rules into the rule files as the current appendRule method does not provide enough flexibility (cant append at the beginning of a rule file or chose a rule file to append to).

Please, if anybody wants to finish this PR feel free to do so.

PR #481, opened 2021-07-04
Exception.hpp: fix build on musl

ffontaine:
Fix the following build failure on musl:

```
In file included from ./src/Library/public/usbguard/Rule.hpp:24,
                 from src/CLI/usbguard-rule-parser.cpp:23:
./src/Library/public/usbguard/Exception.hpp: In static member function 'static std::string usbguard::ErrnoException::reasonFromErrno(int)':
./src/Library/public/usbguard/Exception.hpp:129:72: error: no matching function for call to 'std::__cxx11::basic_string<char>::basic_string(int)'
  129 |       return std::string(strerror_r(errno_value, buffer, sizeof buffer));
      |
```

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/Library/public/usbguard/Exception.hpp
+++ b/src/Library/public/usbguard/Exception.hpp
@@ -126,7 +126,8 @@ namespace usbguard
     static std::string reasonFromErrno(const int errno_value)
     {
       char buffer[1024];
-      return std::string(strerror_r(errno_value, buffer, sizeof buffer));
+      strerror_r(errno_value, buffer, sizeof buffer);
+      return std::string(buffer);
     }
   };
 


ZoltanFridrich:
The problem is that GNU-specific strerror_r() might return some static sring in which case buffer is unused and returning std::string(buffer) will not make much sense. But it definitely does fix the compilation problem with XSI-compliant strerror_r().

ZoltanFridrich (approved):

radosroka:
Can you rebase and force push the PR branch? I'm not able to rerun CI.

ffontaine:
Force push done

radosroka:
@ffontaine we need to support both XSI-compliant and GNU-specific possibilities.

I see from `https://man7.org/linux/man-pages/man3/strerror.3.html`.
```
       strerror_r():
           The XSI-compliant version is provided if:
               (_POSIX_C_SOURCE >= 200112L) && ! _GNU_SOURCE
           Otherwise, the GNU-specific version is provided.
```

You can create conditional `#ifdef` branches for XSI-compliant and GNU-specific implementations.

radosroka:
Can you fix it ?

hartwork:
> Can you fix it ?

@radosroka maybe I can. I have just pushed an alternative PR with a hybrid solution and preprocessor bits as a new pull request #510.

radosroka:
#510 was merged so I think we can close this.

PR #483, opened 2021-07-25
configure.ac: Use standard --with-bash-completion-dir option

thesamesam:
Rationale:
This makes it a lot easier to enable/disable installation
of bash completion files rather than it being based on
a certain package being installed (or not). It's useful
in Gentoo Linux for example where we may want to always
install Bash completion files to the right location, even
if the user isn't using it yet, to save rebuilds.

This makes a few changes to the current Bash completion logic, mainly:
* Falls back gracefully to a standard directory if bash-completion
  itself is not installed (in that case, we can't ask it via pkgconfig
  where to place files).

* So the behaviour is now:
  * no argument / --with-bash-completion-dir=yes:

    Asks pkgconfig, but falls back to standard directory.

  * --with-bash-completion-dir=dir:

    Use the given directory with no detection.

  * --without-bash-completion-dir:

    No installation of Bash completion files.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -438,7 +438,7 @@ usbguard_LDADD=\
 	$(top_builddir)/libusbguard.la \
 	${PTHREAD_LIBS}
 
-if BASH_COMPLETION_ENABLED
+if ENABLE_BASH_COMPLETION
 bashcompletiondir = $(BASH_COMPLETION_DIR)
 dist_bashcompletion_DATA = $(top_srcdir)/scripts/bash_completion/usbguard
 endif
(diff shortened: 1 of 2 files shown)


Cropi:
Hello,

I see your point here and you are definitely right. It's far more cleaner to have bash-completion parameterized rather than depend on if a certain module check fails or not.
My only concern is what happens when you use ```--with-bash-completion-dir=yes``` but you don't have bash-completion installed. Maybe _configure phase_ should error out? I haven't tried it yet.

Anyway, it looks good to me. Thanks for the PR.

Cropi (approved):

PR #492, opened 2021-09-21
Validate ACL permission existence

ZoltanFridrich:
Fail when a valid privilege string is given to a section that does not define behaviour for such privilege.
Example: usbguard add-user testuser -p listen
Privilege listen is irrelevant for section Policy and this call should fail.


Cropi (approved):
Privileges, which do not correspond to a certain section should fail. 
Thanks for the PR.

Issue #460, opened 2021-03-08
usbguard hardening


glitsj16:
To be clear, this is _not_ a bug report, more of _logic check_ or a _feature-request_.

I've been using usbguard (manually built from git) for quite some time, both on Arch Linux and Ubuntu LTS machines. In an attempt to harden my systemd services as tight as possible I noticed usbguard-daemon runs as root and the usbguard.service uses a PIDFile. Both are not optimal IMO (I am aware of the -C switch though).

Recently systemd began [stressing](https://github.com/systemd/systemd/commit/5cfbe7452f023fc307374f306e26b5ee902889f0) that PID files should be avoided in modern projects, advising to use Type=notify or Type=simple to avoid needless forking. Running usbgard `without` the PIDFile=/run/usbguard.pid and as `Type=simple` _seems_ to work just fine when using `Execstart=/usr/bin/usbguard-daemon -s -c /etc/usbguard/usbguard-daemon.conf` (so dropping -f). What is the rationale behind using Type=forking currently?

Trying to run usbguard-daemon as a unprivileged, non-root user seems _impossible_ with the current code base. systemd's `DynamicUser` concept or the more traditional /usr/lib/sysusers.d/foo.conf mechanism both fail due to what I think are the in-built permissions/ownership checks on dirs/files under /etc/usbguard. I sure don't want to use the -P switch to drop these checks, but I've tried and even then the service goes into failure mode. Is there a known, documented way of running usbguard-daemon as unprivileged user (e.g. usbguard)? If not, is such a feature planned or does it make no sense to even consider such a scenario?

radosroka:
Here is the commit a2f89d6edcb4a3d09d5108ef6ca03d88e2971713 that changed type from`simple` to `forking`.

- with such configuration "systemctl start usbguard" can return error code in case of bad configuration

 That is the reason of the change.

glitsj16:
@radosroka Thanks for pointing that out, appreciated. I'll do some more experimenting but I think it's best to close this for now.

PR #498, opened 2021-11-20
Write PIDFile inside /var/run/usbguard directory

juspence:
Related to https://github.com/USBGuard/usbguard/issues/460, it is possible to run USBGuard as an unprivileged (non-root) user even with the -f option & Type=forking in the unit file. To get this to work, I had to:

1) Add a new "usbguard" user and group using systemd-sysusers.
2) Change ownership of "/etc/usbguard/" and "/var/log/usbguard/" to the "usbguard:usbguard" user and group. I may have changed other folders as well but I don't fully remember.
3) Recompile USBGuard to write its PID file to the "/var/run/usbguard/" directory (which can be owned by "usbguard:usbguard") instead of "/var/run/" (which must be owned by root, so writing the PID file as an unprivileged user fails).
4) Update the "PIDFile=/var/run/usbguard/usbguard.pid" option, and add the "User=usbguard", "Group=usbguard", and "SupplementaryGroups=" options to the unit file.

I can't figure out how to 1) create the usbguard user and 2) set the permissions above as part of an RPM install. Is that logic kept in a different repo, or handled by the RPM maintainer?

If that change can be made in this repo, please let me know which file to edit, and I can submit another PR to run USBGuard as a dedicated user. I also need to test a bit more (i.e. reboot) to make sure everything works correctly and there are no errors.


hartwork (commented):
@juspence does the USBGuard code create folder `/var/run/usbguard` anywhere?

juspence:
@hartwork Technically it's Systemd that creates this directory, but it does happen automatically when needed. The line "RuntimeDirectory=usbguard" in the unit file means "create /var/run/usbguard before the service starts".

hartwork:
@juspence thanks for your reply. I think that means that all distros without systemd for an init system will have to extend their init script to create that directory on demand then.

Cropi:
Does the PID file get deleted when the daemon exits due to ```RuntimeDirectory=usbguard```? I am not sure if we do such a cleanup right now, but that could be useful as well.

PR #502, opened 2022-01-14
enhance the fix of race condition while reading connect_type

### LINKED ISSUE #493: race-condition while accessing port/connect_type
On Raspberry Pi, I have observed a race condition in accessing port/connect_type information before symlink is created. Thereby blocking the devices with empty "" connect_type by evaluating the rule `allow with-connect-type "hardwired"`

> Sep 21 07:08:53 raspberrypi4-64 kernel[231]: [   75.929943] hub.c: AFTER device_add
> Sep 21 07:08:53 raspberrypi4-64 usbguard-daemon[249]: [1632208133.175] (T) SysFSDevice.cpp@165/readAttribute: name= idProduct
> Sep 21 07:08:53 raspberrypi4-64 usbguard-daemon[249]: [1632208133.175] (T) SysFSDevice.cpp@165/readAttribute: name= serial
> Sep 21 07:08:53 raspberrypi4-64 usbguard-daemon[249]: [1632208133.176] (T) SysFSDevice.cpp@165/readAttribute: name= authorized
> Sep 21 07:08:53 raspberrypi4-64 usbguard-daemon[249]: [1632208133.176] (T) SysFSDevice.cpp@165/readAttribute: name= port/connect_type
> Sep 21 07:08:53 raspberrypi4-64 kernel: hub.c: AFTER sysfs_create_link for PORT and port/connect_type is created
> Sep 21 07:08:53 raspberrypi4-64 kernel: hub.c: BEFORE sysfs_create_link for DEVICE

From the above log, USBGuard is reading the attributes from SysFs before the symbolic link for port/connect_type

anudeepthika:
Fixes #493 - Hardware and Kernel timings varies on different devices
and while-loop alone doesn't provide considerable delay.
Add small delay in between retry attempts

Host controllers doesn't have port/connect_type
Daemon waits for the entire loop time to return an empty
value. Filter the host controllers and return an empty
string without going through the loop

### DIFF
--- a/src/Library/DeviceBase.cpp
+++ b/src/Library/DeviceBase.cpp
@@ -24,6 +24,7 @@
 #include "DeviceManagerBase.hpp"
 #include "SysFSDevice.hpp"
 #include "Common/FDInputStream.hpp"
+#include "Common/Utility.hpp"
 
 #include "usbguard/Exception.hpp"
 #include "usbguard/Logger.hpp"
@@ -91,19 +92,37 @@ namespace usbguard
      */
     int retry = 0;
     std::string connect_type;
+    /*
+     * Host controllers are directly connected to SoC/PCI bus
+     * hence do not have port/connect_type in sysfs.
+     * Filter out host controllers to avoid unnecessary wait in the loop.
+     */
+    const auto before = std::chrono::steady_clock::now();
 
-    while (retry < 20) {
-      connect_type = sysfs_device.readAttribute("port/connect_type", /*strip_last_null=*/true, /*optional=*/true);
+    if (!hasPrefix(getPort(), "usb")) {
+      while (retry < 30) {
+        connect_type = sysfs_device.readAttribute("port/connect_type", /*strip_last_null=*/true, /*optional=*/true);
 
-      if (connect_type != "") {
-        break;
+        if (connect_type != "") {
+          break;
+        }
+
+        USBGUARD_LOG(Trace) << "connect_type is empty, will attempt again";
+        usleep(1000);
+        retry++;
       }
 
-      USBGUARD_LOG(Trace) << "connect_type is empty, will attempt again";
-      retry++;
+      const auto after = std::chrono::steady_clock::now();
+      const auto duration_nanoseconds = std::chrono::duration_cast<std::chrono::nanoseconds>(after - before).count();
+      USBGUARD_LOG(Info) << "***Setting connect type=" << connect_type <<" after "<< retry << " retries (took " <<
+        duration_nanoseconds << "ns)";
+
+      if (connect_type == "") {
+        USBGUARD_LOG(Warning) << "port/connect_type is empty for non host controller device " << getPort() << " after waiting for " <<
+          duration_nanoseconds << "ns";
+      }
     }
 
-    USBGUARD_LOG(Info) << "***Setting connect type=" << connect_type <<" after retry"<< retry;
     setConnectType(connect_type);
     /*
      * Process USB descriptor data.
--- a/src/Tests/UseCase/003_cli_devices_umockdev.sh
+++ b/src/Tests/UseCase/003_cli_devices_umockdev.sh
@@ -53,6 +53,7 @@ function test_cli_devices()
 
   ${USBGUARD} block-device "$id_dock"
   ${USBGUARD} allow-device "$id_dock"
+  sleep 1
 
   local id_dock_child="$(${USBGUARD} list-devices | sed -n 's|^\([0-9]\+\):.*hash "D3deklX12Ir3kJPfUZ5AQNwHeZn1bwtPkQkw6e+8B38=".*$|\1|p')"
   
@@ -64,6 +65,7 @@ function test_cli_devices()
   ${USBGUARD} block-device "$id_dock_child"
   ${USBGUARD} block-device "$id_dock"
   ${USBGUARD} allow-device "$id_dock"
+  sleep 1
 
   local id_dock_child="$(${USBGUARD} list-devices | sed -n 's|^\([0-9]\+\):.*hash "D3deklX12Ir3kJPfUZ5AQNwHeZn1bwtPkQkw6e+8B38=".*$|\1|p')"
   
--- a/src/Tests/UseCase/003_cli_devices_umockdev.sh
+++ b/src/Tests/UseCase/003_cli_devices_umockdev.sh
@@ -46,7 +46,7 @@ function test_cli_devices()
   local id_dock="$(${USBGUARD} list-devices | sed -n 's|^\([0-9]\+\):.*hash "2y2qS3rcuMr1Ye5knWsbD8CGzPtrs+eiRR/haro7+Ng=".*$|\1|p')"
 
   if [ -z "$id_dock" ]; then
-    echo "Test error: Unable to parse device ID"
+    echo "Test error: Unable to parse device ID (test line ${LINENO})"
     exit 1
   fi
 
@@ -57,7 +57,7 @@ function test_cli_devices()
   local id_dock_child="$(${USBGUARD} list-devices | sed -n 's|^\([0-9]\+\):.*hash "D3deklX12Ir3kJPfUZ5AQNwHeZn1bwtPkQkw6e+8B38=".*$|\1|p')"
   
   if [ -z "$id_dock_child" ]; then
-    echo "Test error: Unable to parse device ID"
+    echo "Test error: Unable to parse device ID (test line ${LINENO})"
     exit 1
   fi
 
@@ -68,7 +68,7 @@ function test_cli_devices()
   local id_dock_child="$(${USBGUARD} list-devices | sed -n 's|^\([0-9]\+\):.*hash "D3deklX12Ir3kJPfUZ5AQNwHeZn1bwtPkQkw6e+8B38=".*$|\1|p')"
   
   if [ -z "$id_dock_child" ]; then
-    echo "Test error: Unable to parse device ID"
+    echo "Test error: Unable to parse device ID (test line ${LINENO})"
     exit 1
   fi
 


anudeepthika:
Hi, @Cropi 

Could you please review this pull request

Cropi:
Hi @anudeepthika ,

Thank you for the pull request. I haven't faced this problem, so I am just a little bit curious if the proposed changes are reliable on your system. Is it enough to wait 10*200 ns? Otherwise, the results will be the same.

anudeepthika:
> Hi @anudeepthika ,
> 
> Thank you for the pull request. I haven't faced this problem, so I am just a little bit curious if the proposed changes are reliable on your system. Is it enough to wait 10*200 ns? Otherwise, the results will be the same.

Hi, @Cropi

Thanks for the response. The issue is easily seen on Raspberry Pi. The wait introduced is 10*200 usec (micro seconds). The total delay of 2 ms (2 milli seconds).

hartwork:
Hi @anudeepthika and @Cropi,

for someone to reproduce the issue scenario in real life and play with the pull request: if they have a Raspberry Pi, one external USB HUB (a host controller?) and one non-HUB USB device at hand, what would they do with the hardware and USBGuard to see the issue in action?

From a code point of view, I understand that a loop without waiting only works if some thing in the loop body is blocking, and hence acting as sleep as a side effect in a way.  Is anything currently blocking in the loop body?

Regarding the sleep duration, I think both "are we sleeping enough?" and "are we sleeping too much?" are valid questions and concerns.  I think maybe there is some maximum waiting time that can be agreed on and then the loop is run not a fixed number of times but until (success or) a monotonic timer (`clock_gettime` with `CLOCK_MONOTONIC`?) has reached that timeout. The overall loop runtime would then be more predictable.

What do you think?

Best, Sebastian

anudeepthika:
> Is anything currently blocking in the loop body



> Hi @anudeepthika and @Cropi,
> 
> for someone to reproduce the issue scenario in real life and play with the pull request: if they have a Raspberry Pi, one external USB HUB (a host controller?) and one non-HUB USB device at hand, what would they do with the hardware and USBGuard to see the issue in action?
> 
> From a code point of view, I understand that a loop without waiting only works if some thing in the loop body is blocking, and hence acting as sleep as a side effect in a way. Is anything currently blocking in the loop body?
> 
> Regarding the sleep duration, I think both "are we sleeping enough?" and "are we sleeping too much?" are valid questions and concerns. I think maybe there is some maximum waiting time that can be agreed on and then the loop is run not a fixed number of times but until (success or) a monotonic timer (`clock_gettime` with `CLOCK_MONOTONIC`?) has reached that timeout. The overall loop runtime would then be more predictable.
> 
> What do you think?
> 
> Best, Sebastian

Hi, @hartwork

Thank you for your response on this PR.

1. To reproduce the issue: You can plug-in a USB device into the Raspberry Pi's USB port and observe the connect_type attribute. Sharing below an observation that I have captured with the race-condition. with-connect-type is supposed to be "unknown" for device with id:7 which is "" here. The linux kernel version is 5.10.31.
Please refer to [this](https://github.com/USBGuard/usbguard/issues/493) issue for the root cause of this race-condition
image
2. Is anything currently blocking in the loop body: Loop without wait is not blocking any other process. Sorry, I couldn't understand this question completely. Could you please elaborate it in detail?

Thanks,
Anu

hartwork:
**Attachment** (renamed git-am patch): [pull-502-play.patch.txt](https://github.com/USBGuard/usbguard/files/7961902/pull-502-play.patch.txt)

---

Hi @anudeepthika thanks for your reply!

I have played with the code myself a bit now to see what durations and retry counts I get over here. With the attached patch applied, I can observe this behavior:

```
# sudo ./usbguard-daemon -k -d |& fgrep took
[1643404384.826] (i) ***Setting connect type=unknown after 0 retries (took 1270ns)
[1643404384.830] (i) ***Setting connect type=unknown after 0 retries (took 636ns)
[1643404384.834] (i) ***Setting connect type=hardwired after 0 retries (took 49148ns)
[1643404384.840] (i) ***Setting connect type=hardwired after 0 retries (took 43271ns)
[1643404384.846] (i) ***Setting connect type=hotplug after 0 retries (took 43353ns)
[1643404384.853] (i) ***Setting connect type=hardwired after 0 retries (took 42260ns)
[1643404384.862] (i) ***Setting connect type=hardwired after 0 retries (took 37831ns)
[1643404384.872] (i) ***Setting connect type=hotplug after 0 retries (took 49301ns)
[1643404384.881] (i) ***Setting connect type=unknown after 0 retries (took 25245ns)
[1643404384.889] (i) ***Setting connect type=unknown after 0 retries (took 23650ns)
[1643404384.898] (i) ***Setting connect type=unknown after 0 retries (took 30134ns)
[1643404388.757] (i) ***Setting connect type=unknown after 0 retries (took 36805ns)
[1643404393.879] (i) ***Setting connect type=unknown after 0 retries (took 38965ns)
```
Note that I'm passing `-k -d` to see debug output, that all retry counts are 0 and that all durations are below 50000ns (just to summarize). I would ask you to apply my attached patch and do the same so that we can get feel for the numbers on that hardware of yours. 

PS: I have used command `for i in $(find /sys/ -name connect_type 2>/dev/null | sort) ; do echo "$i: $(cat $i)" ; done` to see what Linux thinks are the connect types. Based on that I believe it should be "unkown" on timeout and never the empty string.

I'm looking forward to your reply, best, Sebastian

anudeepthika:
> **Attachment** (renamed git-am patch): [pull-502-play.patch.txt](https://github.com/USBGuard/usbguard/files/7961902/pull-502-play.patch.txt)
> 
> Hi @anudeepthika thanks for your reply!
> 
> I have played with the code myself a bit now to see what durations and retry counts I get over here in practice. With the attached patch applied, I can observe this behavior:
> 
> ```
> # sudo ./usbguard-daemon -k -d |& fgrep took
> [1643404384.826] (i) ***Setting connect type=unknown after 0 retries (took 1270ns)
> [1643404384.830] (i) ***Setting connect type=unknown after 0 retries (took 636ns)
> [1643404384.834] (i) ***Setting connect type=hardwired after 0 retries (took 49148ns)
> [1643404384.840] (i) ***Setting connect type=hardwired after 0 retries (took 43271ns)
> [1643404384.846] (i) ***Setting connect type=hotplug after 0 retries (took 43353ns)
> [1643404384.853] (i) ***Setting connect type=hardwired after 0 retries (took 42260ns)
> [1643404384.862] (i) ***Setting connect type=hardwired after 0 retries (took 37831ns)
> [1643404384.872] (i) ***Setting connect type=hotplug after 0 retries (took 49301ns)
> [1643404384.881] (i) ***Setting connect type=unknown after 0 retries (took 25245ns)
> [1643404384.889] (i) ***Setting connect type=unknown after 0 retries (took 23650ns)
> [1643404384.898] (i) ***Setting connect type=unknown after 0 retries (took 30134ns)
> [1643404388.757] (i) ***Setting connect type=unknown after 0 retries (took 36805ns)
> [1643404393.879] (i) ***Setting connect type=unknown after 0 retries (took 38965ns)
> ```
> 
> Note that I'm passing `-k -d` to see debug output, that all retry counts are 0 and that all durations are below 50000ns (just to summarize). I would ask you to apply my attached patch and do the same so that everyone can get a better feel for the numbers on that hardware of yours.
> 
> PS: I have used command `for i in $(find /sys/ -name connect_type 2>/dev/null | sort) ; do echo "$i: $(cat $i)" ; done` to see what Linux thinks are the connect types. Based on that I believe it should be "unkown" on timeout and never the empty string.
> 
> I'm looking forward to your reply, best, Sebastian


Hi @hartwork 

Thanks for sending the patch. I have applied the patch and compiled it on my hardware. Below are the observations captured. The first two entries are host controllers, 3rd entry is the root hub which I have marked as "hardwired" and starting from 4th entry are USB devices that I plugged into various USB ports on the root hub ( 2 are hardwired and 2 are default (unknown))
image

The maximum retry number is 40 and all durations are more than 50000ns except for the first two entries which are host controllers.

The patch that you've sent passes the default connect_type as "unknown" but host controllers don't have the port/connect_type attribute and hence should be empty.

Returning "unknown" instead of an empty string on retry failure would still block the USB devices plugged into "hardwired or hotplug" ports when there is a rule to allow the devices with-connect-type as "hardwired" or "hotplug" and the user might misinterpret the connect_type as "unknown".

Sorry for the delay in response.

Regards,
Anu

hartwork:
> I'm still wondering if we should use (a) a fixed retry count or (b) just look at the spent duration for leaving the loop. Former is easier to implement and may use minimally less CPU, latter seems more predictable. If we have the time tracking for better messages already, the gap towards using it for loop termination is smaller. What do you think? @Cropi how about you?

Hi @anudeepthika, I thought about this^^ more: If we quit the loop on timeout, it will be the same on slow and fast hardware. If we break after 30 retries, slow hardware will get slightly get more time because code runs slower in general. That is why I think maybe sticking to terminating after n retries is not just slightly simpler to implement (and with a smaller diff) but may also help slow hardware without hurting fast one. If you're undecided, let's go with that. If you want the opposite, also good with me, it's not a huge difference.

With that out of the way I will do "regular code review" on the current diff now to get things rolling.
Personally, I think it would be great to get this PR ready and merged within say 7 days so that a new release with this fixed becomes possible (#504). Would that work for your schedule?

Best, Sebastian

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
      * Set connect type
      */
     int retry = 0;
-    std::string connect_type;
-    while (retry < 20) {
+    std::string connect_type = "";
+    /*
+     * Read port/connect_type only for non Host Controller devices
+     */
+    if (getPort().substr(0, 3) != "usb") {
```
hartwork, line 97: My vote for turning this into
- a) `hasPrefix(getPort(), "usb")` plus `#include "Common/Utility.hpp"`
- b) `isController()` (but but does more than just checking the port, please double check if that's in our interest, I'm unsure)

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
      * Set connect type
      */
     int retry = 0;
-    std::string connect_type;
-    while (retry < 20) {
+    std::string connect_type = "";
+    /*
+     * Read port/connect_type only for non Host Controller devices
+     */
```
hartwork, line 96: If it can be explained in a few words *why* a host controller does not have a connect type, I think that may add value. Just an idea.

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
      * Set connect type
      */
     int retry = 0;
-    std::string connect_type;
-    while (retry < 20) {
+    std::string connect_type = "";
```
hartwork, line 93: Adding ` = ""` does not seem to change semantics or add value. I checked back with http://www.cplusplus.com/reference/string/string/string/. Let's drop that part for a cleaner diff, okay?

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
...
+    /*
+     * Read port/connect_type only for non Host Controller devices
+     */
+    if (getPort().substr(0, 3) != "usb") {
+      while (retry < 10) {
```
hartwork, line 98: Based on your screeenshot with real life Pi durations, I suggest `retry < 30` (and `usleep(1000 * 1000);` below) to sleep up to 30 milliseconds total.

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
...
         connect_type = sysfs_device.readAttribute("port/connect_type", /*strip_last_null=*/true, /*optional=*/true);
         if (connect_type != "") {
             break;
         }
         USBGUARD_LOG(Trace) << "connect_type is empty, will attempt again";
+        usleep(200);
```
hartwork, line 104: Please see my combined comment above :point_up:

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
...
         retry++;
+      }
```
hartwork, line 106: Please note that on target branch `master` formatting of file `src/Library/DeviceBase.cpp` has changed in the mean time (in commit be5d74bd34458b9c02ff352599b73a34417e3c3e) and we also gained additional CI in the mean time that should get a say in this PR prior to a merge. Please rebase your PR onto latest `master`, fix the unavoidable conflicts, use script `./scripts/reformat-sources.sh` to ensure formatting that the CI will have a chance to like, and then force-push back into the same pull request. Thank you!

hartwork (changes requested):

src/Library/DeviceBase.cpp
```diff
@@ -90,14 +90,20 @@ namespace usbguard
...
     }
     USBGUARD_LOG(Info) << "***Setting connect type=" << connect_type <<" after retry"<< retry;
```
hartwork, line 108: I forgot: It would be great to have this output line mention how long things took, e.g. so that user can proof based on logs whenever the current values are not sufficient, in the future. Consider my earlier demo code doing something like that CC-0 ("public domain") in case you want to copy from there.

anudeepthika:
> > I'm still wondering if we should use (a) a fixed retry count or (b) just look at the spent duration for leaving the loop. Former is easier to implement and may use minimally less CPU, latter seems more predictable. If we have the time tracking for better messages already, the gap towards using it for loop termination is smaller. What do you think? @Cropi how about you?
> 
> Hi @anudeepthika, I thought about this^^ more: If we quit the loop on timeout, it will be the same on slow and fast hardware. If we break after 30 retries, slow hardware will get slightly get more time because code runs slower in general. That is why I think maybe sticking to terminating after n retries is not just slightly simpler to implement (and with a smaller diff) but may also help slow hardware without hurting fast one. If you're undecided, let's go with that. If you want the opposite, also good with me, it's not a huge difference.
> 
> With that out of the way I will do "regular code review" on the current diff now to get things rolling. Personally, I think it would be great to get this PR ready and merged within say 7 days so that a new release with this fixed becomes possible (#504). Would that work for your schedule?
> 
> Best, Sebastian

Hi @hartwork,

Thanks for the comment. I have decided to go for a fixed retry plus sleep approach since it helps slow hardware without hurting fast one (like you said). I agree with the above calculations and will implement the same.

Thanks for the review comments. I will resolve and push the latest changes at the earliest.

Regards,
Anu

hartwork:
@anudeepthika awesome, I'm looking forward to the next version of this pull request :+1:

src/Library/DeviceBase.cpp
anudeepthika: I would go with option "a"

I have tried using isController() method but the additional information that it is checking more than port is not available in the sysfs at this point of code. The descriptors are parsed later the connect_type is set and host controllers can't be filtered using isController() here. That additional check is out of our interest here.

src/Library/DeviceBase.cpp
anudeepthika: Yes, I will update retry to 30 and usleep to 1000 (not 1000 * 1000) since usleep()  suspends execution for microsecond intervals (not nanoseconds).

src/Library/DeviceBase.cpp
hartwork: Sounds good!

src/Library/DeviceBase.cpp
hartwork: My bad, good point! :+1:

anudeepthika:
Hi @hartwork,

Please find the attached patch with the changes suggested in the review comments.
[0001-enhance-the-fix-of-race-condition-while-reading-conn.patch.txt](https://github.com/USBGuard/usbguard/files/8003206/0001-enhance-the-fix-of-race-condition-while-reading-conn.patch.txt)

I have faced a `make check` failure with this patch. The test case case [003_cli_devices_umockdev.sh](https://github.com/USBGuard/usbguard/blob/master/src/Tests/UseCase/003_cli_devices_umockdev.sh) has failed check. The [mock data](https://github.com/USBGuard/usbguard/blob/master/src/Tests/UseCase/devices.umockdev) provided to this test case doesn't have port/connect_type information hence going through complete loop time. Looks like there is a defined timeout for these tests (around 3 msec, I'm unsure) exceeding that time is failing the check.

Could you please update the mock data appropriately so that this patch passes all the tests?

anudeepthika:
> @anudeepthika I'll have a look now and see what I can do. Any reason why you're not putting it into the pull request? The sooner we get full CI looking at, the better. If it starts out red, that's okay.

Sure, thank you. I'll push the changes to the PR soon.

Regards,
Anu

hartwork:
@anudeepthika I had a chance to play with the patch now:

> Please find the attached patch with the changes suggested in the review comments. [0001-enhance-the-fix-of-race-condition-while-reading-conn.patch.txt](https://github.com/USBGuard/usbguard/files/8003206/0001-enhance-the-fix-of-race-condition-while-reading-conn.patch.txt)

Looks pretty good, nice C++ style monotonic clock. Maybe add one (or two?) full stop(s) to the comment, but all else is great already :+1: 

> I have faced a `make check` failure with this patch. The test case case [003_cli_devices_umockdev.sh](https://github.com/USBGuard/usbguard/blob/master/src/Tests/UseCase/003_cli_devices_umockdev.sh) has failed check. The [mock data](https://github.com/USBGuard/usbguard/blob/master/src/Tests/UseCase/devices.umockdev) provided to this test case doesn't have port/connect_type information hence going through complete loop time. Looks like there is a defined timeout for these tests (around 3 msec, I'm unsure) exceeding that time is failing the check.
> 
> Could you please update the mock data appropriately so that this patch passes all the tests?

There are two places where I needed to add `sleep 1` in `003_cli_devices_umockdev.sh` to make things work. I also adjusted the test to help the reader understanding which of the three `Test error: Unable to parse device ID` is speaking, if later changes turn out to need more than 1 second of rest. It help me while figuring things out. So I have added your commit and these two commits of mine and then the full CI is green for me:
- [`0001-tests-Make-003_cli_devices_umockdev.sh-error-output-.patch`](https://github.com/USBGuard/usbguard/files/8006020/0001-tests-Make-003_cli_devices_umockdev.sh-error-output-.patch.txt)
- [`0002-tests-Fix-003_cli_devices_umockdev.sh-by-adding-mini.patch`](https://github.com/USBGuard/usbguard/files/8006023/0002-tests-Fix-003_cli_devices_umockdev.sh-by-adding-mini.patch.txt)

Best, Sebastian

hartwork:
PS: @anudeepthika could you put `Fixes #493` in the first line of the topmost description (where it say "Hardware and Kernel [..]")? That should get the ticket "linked" and have it be auto-closed on merge.

anudeepthika:
> @anudeepthika I had a chance to play with the patch now:
> 
> > Please find the attached patch with the changes suggested in the review comments. [0001-enhance-the-fix-of-race-condition-while-reading-conn.patch.txt](https://github.com/USBGuard/usbguard/files/8003206/0001-enhance-the-fix-of-race-condition-while-reading-conn.patch.txt)
> 
> Looks pretty good, nice C++ style monotonic clock. Maybe add one (or two?) full stop(s) to the comment, but all else is great already 👍

Thanks for the review. I have added full stops and pushed the commit.
> 
> > I have faced a `make check` failure with this patch. The test case case [003_cli_devices_umockdev.sh](https://github.com/USBGuard/usbguard/blob/master/src/Tests/UseCase/003_cli_devices_umockdev.sh) has failed check. The [mock data](https://github.com/USBGuard/usbguard/blob/master/src/Tests/UseCase/devices.umockdev) provided to this test case doesn't have port/connect_type information hence going through complete loop time. Looks like there is a defined timeout for these tests (around 3 msec, I'm unsure) exceeding that time is failing the check.
> > Could you please update the mock data appropriately so that this patch passes all the tests?
> 
> There are two places where I needed to add `sleep 1` in `003_cli_devices_umockdev.sh` to make things work. I also adjusted the test to help the reader understanding which of the three `Test error: Unable to parse device ID` is speaking, if later changes turn out to need more than 1 second of rest. It help me while figuring things out. So I have added your commit and these two commits of mine and then the full CI is green for me:
> 
> * [`0001-tests-Make-003_cli_devices_umockdev.sh-error-output-.patch`](https://github.com/USBGuard/usbguard/files/8006020/0001-tests-Make-003_cli_devices_umockdev.sh-error-output-.patch.txt)
> * [`0002-tests-Fix-003_cli_devices_umockdev.sh-by-adding-mini.patch`](https://github.com/USBGuard/usbguard/files/8006023/0002-tests-Fix-003_cli_devices_umockdev.sh-by-adding-mini.patch.txt)
> 
> Best, Sebastian

Hi @hartwork,

Thanks for the patches. All the checks have passed with these patches applied. Do you want me to push those patches (with commit author information retained) as part of this PR?

Regards,
Anu

hartwork:
> Thanks for the review. I have added full stops and pushed the commit.

@anudeepthika excellent, thank you! :+1: 


> Thanks for the patches. I've got all the checks passed on my system with these patches applied. Do you want me to push those patches (with commit author information retained) as part of this PR?

I'm open to other approaches but `git am`'ing them into this PR would definitely work for me.

Best, Sebastian

hartwork:
@anudeepthika PS: I should mention that the current LGTM complaints will need PR #528 to go away.

anudeepthika:
> > Thanks for the review. I have added full stops and pushed the commit.
> 
> @anudeepthika excellent, thank you! 👍
> 
> > Thanks for the patches. I've got all the checks passed on my system with these patches applied. Do you want me to push those patches (with commit author information retained) as part of this PR?
> 
> I'm open to other approaches but `git am`'ing them into this PR would definitely work for me.
> 
> Best, Sebastian

Sure, I'll do that and push those commits. Thank you.

Regards,
Anu

anudeepthika:
> @anudeepthika PS: I should mention that the current LGTM complaints will need PR #528 to go away.

Sure, thank you.

hartwork (dismissed):
:+1: Good to merge in my eyes, @radosroka what do you think?

anudeepthika:
Hi @hartwork,

I have made a minor change in my commit. I have added a warning when connect_type is empty after the wait time. This will make the user notice the warning without enabling debug mode.

Please provide your comments.

Regards,
Anu

hartwork:
> I have made a minor change in my commit. I have added a warning when connect_type is empty after the wait time. This will make the user notice the warning without enabling debug mode.

@anudeepthika I think that's a good idea. Maybe the details are not "perfect" yet, e.g. for the user to act upon it they would probably need to know which device it's about and the duration (not sure about retry count). Also for host controllers it's fine so the message could go further saying that it's a non-controller or why or something. I'd also flip the order, to have low-level the debug one go first, which I think will be more natural to some with debug output enabled.

anudeepthika:
> > I have made a minor change in my commit. I have added a warning when connect_type is empty after the wait time. This will make the user notice the warning without enabling debug mode.
> 
> @anudeepthika I think that's a good idea. Maybe the details are not "perfect" yet, e.g. for the user to act upon it they would probably need to know which device it's about and the duration (not sure about retry count). Also for host controllers it's fine so the message could go further saying that it's a non-controller or why or something. I'd also flip the order, to have low-level the debug one go first, which I think will be more natural to some with debug output enabled.

Hi @hartwork,

Thanks for the suggestion. I have flipped the order. Low-level debug one first and then warning with device information and duration in the warning message for non host controller devices.

hartwork (approved):

radosroka (approved):
Thank you. I'll merge it.

PR #514, opened 2022-01-28
Fix compilation with musl + make GitHub Actions protect against future musl regressions

hartwork:
It turns out that #510 needed more work and that function `basename` causes similar trouble.
So this pull request fixes both issues and adds related CI coverage, so we won't walk into musl regressions unnoticed.
I hope you like it :beers: 

CC @radosroka

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,6 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         linux_distro:
+        - Alpine Linux 3.15  # with musl
         - CentOS 8.2         # with GCC 8.5.0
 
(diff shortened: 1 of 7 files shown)


radosroka:
As I wrote in #515, separate directory for dockerfiles would be great. Other parts are OK. I would like to merge the #515 first.

hartwork:
@radosroka rebased and pushed now

hartwork:
PS: Please note that the red part of the CI is unrelated to this PR and that the CI being added here is fully green, see `Build on non-Ubuntu Linux using Docker` below. (I can rebase on top of #519 if you'd rather have a fully green CI before merge, as you prefer.)

radosroka (approved):

PR #515, opened 2022-01-29
Fix compilation for PEGTL 3 with GCC 8 + make GitHub Actions cover GCC 8.5.0 (fixes #394)

### LINKED ISSUE #394: Link error due to missing -lstdc++fs (e.g. on CentOS 8.2)
Same error with CentOS 8.2.

Trying to configure with:
./configure --with-crypto-library=sodium --with-bundled-catch --with-bundled-pegtl

Getting error:
....
configure: Using bundled Catch library
checking tao/pegtl.hpp usability... no
checking tao/pegtl.hpp presence... no
checking for tao/pegtl.hpp... no
configure: error: in `/home/username/rpmbuild/BUILD/usbguard-0.7.8':
configure: error: PEGTL header file not found or not usable. Re-run with --with-bundled-pegtl to use the bundled library.

Also tried compiling PEGTL without any success. Official USBGuard release in CentOS8 is not stable. Really need to this.

Please help or point me in any direction!

Best regards

hartwork:
Fixes #394

This fixes the [link errors](https://github.com/USBGuard/usbguard/issues/394#issuecomment-1022530220) that happen with PEGTL 3 and GCC 8 on e.g. CentOS 8.2.
The CI part is the same approach as in PR #514 and I plan to rebase one onto the other, once either of them is merged.
I hope you like this direction, I'm looking forward to your review :beers: 

CC @Cropi @radosroka

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,48 @@
+##
+## Copyright (c) 2022 Sebastian Pipping <sebastian@pipping.org>
+##
+## This program is free software; you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published by
+## the Free Software Foundation; either version 2 of the License, or
+## (at your option) any later version.
+##
+## This program is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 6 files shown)


radosroka (approved):

radosroka:
Actually I would move the docker file to the scripts/dockerfiles directory. 

We can extend the docker.yml with other linux distros later.

hartwork:
> Actually I would move the docker file to the scripts/dockerfiles directory.

@radosroka sure! Let's make it `scripts/docker/`, okay? That seems less limiting in future directory content, still gives the needed information. Adjusted and pushed now.

> We can extend the docker.yml with other linux distros later.

Yes, that was the idea. For anyone else reading, PR #514 is going in exactly that direction.

radosroka (approved):

PR #518, opened 2022-02-02
Fix approach to "make distcheck" in GitHub Actions

hartwork:
- The test log file location for `make check` and `make distcheck` differs and was mixed up:
  - `make check` has `src/Tests/test-suite.log`.
  - `make distcheck` has `usbguard-*/_build/sub/src/Tests/test-suite.log`.

  The new approach unhardcodes the path to avoid regressions, and dumps both if available.

- `make distcheck` has its own set of configure arguments (in variable `AM_DISTCHECK_CONFIGURE_FLAGS`) and so (while of value) was not a good substitute for `make check` (with custom configure arguments). So now both `make check` and `make distcheck` are run.

- Both `make check` nor `make distcheck` and have turned out not to fully robust towards `-j$(nproc)`, so now make option `-j1` is used to make that explicit.

CC @radosroka


radosroka (approved):

radosroka:
Thank you :)

PR #523, opened 2022-02-02
Improve Readme (Ubuntu 21.10 instructions, help with first start, more) + make CI cover Ubuntu 21.10 (fixes #474, fixes #500)

### LINKED ISSUE #474: Installation procedure and doc could be better
In order to make it work on debian 10/ ubuntu 20.04 I had to:
1. install dependencies
    ```
    sudo apt install -y protobuf-compiler libsodium-dev libpolkit-gobject-1-dev libaudit-dev libcap-ng-dev libldap-dev
    ```
    Of course, some of deps were not necessary, but it was far not a first attempt. [CI config](https://github.com/USBGuard/usbguard/blob/master/.github/workflows/build.yml#L87) helped a bunch.
2. configure
    ```
    ./configure --with-bundled-catch --with-bundled-pegtl --enable-systemd
    ```
    and carefully follow the warnings in case if something important was missing.
3. `make` and carefully watch the logs
4. `sudo make install` and carefully watch the logs
5. check where it created the config files, in my case it was `/usr/local/etc/usbguard/`
6. `make check` should not fail
7. **Important: do not [lock](https://gist.github.com/pjobson/2f4979e7570f4f3f9a34e42f5617488a) yourself from your USB devices** 
    generate the current config with
    `sudo usbguard generate-policy > /usr/local/etc/usbguard/rules.conf` 
    thanks [arch wiki](https://wiki.archlinux.org/index.php/USBGuard)
8. `sudo systemctl start usbguard.service`
9. `sudo systemctl status usbguard.service` this shouldn't return errors, if it does, stop, fix and restart
10. `sudo usbguard list-rules` check if important devices are allowed
11. `sudo systemctl enable usbguard.service` when everything works

This is maybe a bit extensive to put into a doc, but there might be something like a `QuickStart.md` for mere mortals.
Please tell me what you think and I'll do a PR.

### LINKED ISSUE #500: Ubuntu install instructions (for 21.10)
Thought I'd leave these here in case others want to install, as it was a multi-step process with plenty of digging for me:

```

sudo apt install protobuf-compiler autoconf libtool libseccomp-dev libaudit-dev libcap-ng-dev libumockdev-dev libdbus-glib-1-dev libxml2-utils xsltproc libpolkit-gobject-1-dev 
./autogen.sh
./configure --with-crypto-library=gcrypt --with-bundled-catch --with-bundled-pegtl --enable-systemd --with-dbus
make
sudo make install
```

After these steps I still wasn't able to run `usbguard` from the `/usr/local/bin` path though, only from the build folder. Issue https://github.com/USBGuard/usbguard/issues/485

hartwork:
Fixes #474 — Installation procedure and doc could be better
Fixes #500 — Ubuntu install instructions (for 21.10)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -28,6 +28,7 @@ jobs:
         linux_distro:
         - Alpine Linux 3.15  # with musl
         - CentOS 8.2         # with GCC 8.5.0
+        - Ubuntu 21.10       # because super popular
 
     name: Build on ${{ matrix.linux_distro }} using Docker
     runs-on: ubuntu-20.04
--- a/README.adoc
+++ b/README.adoc
@@ -17,12 +17,6 @@ image::https://img.shields.io/github/license/USBGuard/usbguard.svg[License, link
 USBGuard is a software framework for implementing USB device authorization policies (what kind of USB devices are authorized) as well as method of use policies (how a USB device may interact with the system).
 Simply put, it is a USB device allowlisting tool.
 
-WARNING: The 0.x releases are not production ready packages.
-They serve for tech-preview and user feedback purposes only.
-Please share your feedback or request a feature in the Github issue trackers for each project:
-
- * https://github.com/USBGuard/usbguard/issues/new[Report a bug or request a feature in *usbguard*]
-
 == Documentation
 
  * User Guide (TBA)
@@ -34,6 +28,14 @@ Please share your feedback or request a feature in the Github issue trackers for
  ** <<doc/man/usbguard-dbus.8.adoc#name, usbguard-dbus(8)>>
 
 == Compilation & Installation
+
+WARNING: *Prior to starting the USBGuard daemon (or service) for the first time*
+         (but after installation)
+         we need to
+         generate a rules file for USBGuard so that the currently attached
+         USB devices (in particular mouse and keyboard) keep working
+         so that you will not **get locked out of your system**.
+         More on that below at <<before-the-first-start, Before the First Start>>.
 
 To compile the source code, you will require at least C{plus}{plus}17. +
 If you are compiling sources from a release tarball, you'll need the development files for:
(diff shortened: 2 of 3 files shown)


radosroka:
Could you create two separate commits, one for README and the second one for docker?

hartwork:
> Could you create two separate commits, one for README and the second one for docker?

@radosroka it's split into 8 commits currently so those two parts are already separate, I don't understand. Do you want me to to split the Docker into ta separate PR? The reason I combined it is because they are strongly related: The readme says what the Dockerfile says, minus package systemd. I was unsure if the CI should spend time on Ubuntu 21.10 or not. I have dropped the Docker part now, pushed.

radosroka:
> > Could you create two separate commits, one for README and the second one for docker?
> 
> @radosroka it's split into 8 commits currently so those two parts are already separate, I don't understand. Do you want me to to split the Docker into ta separate PR? The reason I combined it is because they are strongly related: The readme says what the Dockerfile says, minus package systemd. I was unsure if the CI should spend time on Ubuntu 21.10 or not. I have dropped the Docker part now, pushed.

I just wanted two commits in this PR. One with the README changes(squashed) and the second with docker.

hartwork:
> > > Could you create two separate commits, one for README and the second one for docker?
> > 
> > 
> > @radosroka it's split into 8 commits currently so those two parts are already separate, I don't understand. Do you want me to to split the Docker into ta separate PR? The reason I combined it is because they are strongly related: The readme says what the Dockerfile says, minus package systemd. I was unsure if the CI should spend time on Ubuntu 21.10 or not. I have dropped the Docker part now, pushed.
> 
> I just wanted two commits in this PR. One with the README changes(squashed) and the second with docker.

@radosroka I think individual commits were better but I can do, one moment.

hartwork:
> > I just wanted two commits in this PR. One with the README changes(squashed) and the second with docker.
> 
> @radosroka I think individual commits were better but I can do, one moment.

@radosroka done, pushed, two commits now

radosroka (approved):

PR #531, opened 2022-02-06
[security] Fix unauthorized access via D-Bus (fixes #273, fixes #403)

### LINKED ISSUE #273: No default ACL on some dbus methods
Some of the methods exposed via dbus are not covered by the [current default policy](https://github.com/USBGuard/usbguard/blob/master/src/DBus/org.usbguard.policy). This includes the `getParameter` and `setParameter` methods.

On a default debian install, as unprivileged user:
```
$ usbguard get-parameter InsertedDevicePolicy
ERROR: IPC connect: service=usbguard: Operation not permitted
$ dbus-send --system --print-reply --dest=org.usbguard /org/usbguard/Policy org.usbguard.getParameter string:"InsertedDevicePolicy"
method return time=1549540531.655300 sender=:1.79383 -> destination=:1.120734 serial=97 reply_serial=2
   string "apply-policy"
```

This can be fixed by adding these to the current policy configuration. However, having 2 buses and 2 ACLs makes the administration of any deployment tedious. There is always a risk of future methods to be exposed. Ideally, dbus should be the only bus exposed and the libqb bus dropped.

### LINKED ISSUE #403: Polkit policy is not valid with newest versions of Polkit
With new versions of Polkit, the value `auth_self_keep_session` is no more valid for the tags `<allow_any>`, `<allow_inactive>` and `<allow_active>`.
The [usbguard policy config](https://github.com/USBGuard/usbguard/blob/master/src/DBus/org.usbguard1.policy) is not parsed anymore by new versions of Polkit, and the defined actions are unavailable.

#### proposed solutions

*  `auth_self_keep_session`: allowed executing an action after authenticating, and the redo the action without auth for all the rest of the session.
 It is now unavailable

* `auth_self_keep`: the authorization is kept for a brief period (e.g. five minutes).
It is most probably not retro-compatible. (I found no reference online of `auth_self_keep` on pages mentioning `auth_self_keep_session`)

* `auth_self`: You have to re-authenticate at each call of the action.
 This argument is retro-compatible.

Should we replace occurrences of `auth_self_keep_session` with `auth_self_keep` or `auth_self`?

hartwork:
Fixes #273
Fixes #403

- Fix `.policy` file:
  - Replace unsupported value `auth_self_keep_session` by `auth_self_keep` to fix the Polkit parse error so that Polkit stops ignoring the USBGuard `.policy` file.
  - Drop actions/methods from the policy that no longer exist.
  - Add actions/methods to the policy file that have been missing.
- Make `usbguard-dbus` call out to Polkit for authorization so that the policies from the `.policy` file are actually respected.
- Add CI to protect against policy file parse error regressions.

CC @radosroka @Cropi

### DIFF
--- a/.github/workflows/polkit.yml
+++ b/.github/workflows/polkit.yml
@@ -1,0 +1,67 @@
+##
+## Copyright (c) 2022 Sebastian Pipping <sebastian@pipping.org>
+##
+## This program is free software; you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published by
+## the Free Software Foundation; either version 2 of the License, or
+## (at your option) any later version.
+##
+## This program is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+name: Check for Polkit policy parse errors
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  polkit_policies:
+
+    name: Check for Polkit policy parse errors
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install runtime dependencies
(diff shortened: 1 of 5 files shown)


radosroka (approved):

radosroka:
I don't have any objections.

PR #541, opened 2022-03-05
Restore support for access control filenames without a group (fixes #540)

### LINKED ISSUE #540: Ignoring access control file because of malformed name
Ever since upgrading to usbguard 1.1.0, I'm getting this error `ERROR: IPC connect: service=usbguard: Operation not permitted` when trying to use usbguard as my user. I ran `usbguard-daemon -d` and the logs showed
```
[1646452999.870] (D) Daemon.cpp@262/loadConfiguration: Setting IPCAllowedUsers to { root }
[1646452999.870] (T) Daemon.cpp@1086/addIPCAllowedUser: user=root
[1646452999.870] (D) Daemon.cpp@274/loadConfiguration: Setting IPCAllowedGroups to {  }
[1646452999.870] (D) Daemon.cpp@284/loadConfiguration: Setting DeviceRulesWithPort to false
[1646452999.870] (i) File has correct permissions.
[1646452999.870] (i) Loading IPC access control files at /etc/usbguard/IPCAccessControl.d
[1646452999.870] (T) Utility.cpp@361/loadFiles: L: brady : /etc/usbguard/IPCAccessControl.d/brady
[1646452999.870] (i) Loading IPC access control file /etc/usbguard/IPCAccessControl.d/brady
[1646452999.870] (W) Ignoring access control file because of malformed name: brady
```
The file `/etc/usbguard/IPCAccessControl.d/brady` was created by
```
sudo usbguard add-user brady --policy list --devices list,listen --exceptions listen --parameters list,listen
```
and it has the correct permissions.

I saw #479 and thought maybe that is causing the issue. I'm using Arch Linux in case it matters.

hartwork:
Fixes #540

Regression from commit b15ef713a9ac47e84525bbf829c7f444b84c3c81 of release 1.1.0, detailed analysis online at https://github.com/USBGuard/usbguard/issues/540#issuecomment-1059784284 .


ZoltanFridrich:
@hartwork Please, add the same check for username as well. You can have ACL files in a form ":groupname", ie. with empty username.

hartwork:
> @hartwork Please, add the same check for username as well. You can have ACL files in a form ":groupname", ie. with empty username.

@ZoltanFridrich @Cropi good point, thanks for speaking up about it! I wonder if cases `user:` and `:`  should be (or were previously) supported. I have updated the PR with a version including those two cases and adding a comment about all the effectively supported variations. Please re-check, pushed.

ZoltanFridrich:
LGTM!
I am not sure how it works with `:` name. Does it work like a global IPC ACL meaning that it gets applied for everyone?

hartwork:
> LGTM! I am not sure how it works with `:` name. Does it work like a global IPC ACL meaning that it gets applied for everyone?

That's a good point to bring up! From looking at `IPCServerPrivate::matchACLByName` code, my impression is that we are dealing with algorithm "start with nothing, then match user OR match group" rather than "start with everyone, then cut things away". I believe that filename `:` does effectively match no one (rather than everyone). I'm not sure that's the best way to implement it, but we probably cannot change semantics of it before 2.0.0 if we wanted to. My personal focus would be to support what was supported in 1.0.0. @ZoltanFridrich what do you think?

ZoltanFridrich:
I was just curious about `:` name situation. I agree with you and I think this PR can be merged.

radosroka (approved):
Thank you!

PR #542, opened 2022-03-11
Use authentication instead of authentification

b1rger:
Authentication is the preferred form in English, see https://english.stackexchange.com/questions/7844/is-authentification-a-real-word#7845


hartwork (approved):
Intereresting, I was probably thinking too German ("Authentifikation") with this indeed.

radosroka (approved):

PR #546, opened 2022-04-15
dbus: check whether the client wanted interactive authentication

muelli:
Some clients do not want to prompt for passwords, e.g. when authorising keyboards
when the screen is locked, prompting for a password doesn't make sense because
the user cannot enter anything.
This change extracts the flags and checks whether the client called with
G_DBUS_CALL_FLAGS_ALLOW_INTERACTIVE_AUTHORIZATION.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -467,7 +467,7 @@ if test "x$with_dbus" = xyes; then
   #
   # Check for required D-Bus modules
   #
-  PKG_CHECK_MODULES([dbus], [dbus-1 gio-2.0 polkit-gobject-1],
+  PKG_CHECK_MODULES([dbus], [dbus-1 gio-2.0 >= 2.46 polkit-gobject-1],
   [AC_DEFINE([HAVE_DBUS], [1], [Required GDBus API available])
   dbus_summary="system-wide; $dbus_CFLAGS $dbus_LIBS"],
   [AC_MSG_FAILURE([Required D-Bus modules (dbus-1, gio-2.0) not found!])]
--- a/src/DBus/DBusBridge.cpp
+++ b/src/DBus/DBusBridge.cpp
@@ -477,7 +477,11 @@ namespace usbguard
     polkit_details_insert (details, "polkit.message", "This USBGuard action needs authorization");
     USBGUARD_LOG(Trace) << "Customized.";
     USBGUARD_LOG(Trace) << "Checking authorization of action \"" << action_id.str() << "\" with Polkit ...";
-    const PolkitCheckAuthorizationFlags flags = POLKIT_CHECK_AUTHORIZATION_FLAGS_ALLOW_USER_INTERACTION;
+    GDBusMessage* const message = g_dbus_method_invocation_get_message (invocation);
+    const PolkitCheckAuthorizationFlags flags = (g_dbus_message_get_flags (message) &
+        G_DBUS_MESSAGE_FLAGS_ALLOW_INTERACTIVE_AUTHORIZATION)
+      ? POLKIT_CHECK_AUTHORIZATION_FLAGS_ALLOW_USER_INTERACTION
+      : POLKIT_CHECK_AUTHORIZATION_FLAGS_NONE;
     PolkitAuthorizationResult* const result = polkit_authority_check_authorization_sync
       (authority,
         subject,


muelli:
I'm coming from https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/issues/676 and I haven't been able to test this, yet. Sorry.

muelli:
@hartwork may have an opinion here, too :)

hartwork (changes requested):
Hi @muelli,

that's a nice idea, thanks for bringing this to my attention and for taking it upstream! :+1:
Some notes:

- Flag `G_DBUS_MESSAGE_FLAGS_ALLOW_INTERACTIVE_AUTHORIZATION` seems to require glib >=2.46 according to https://libsoup.org/gio/GDBusMessage.html.
  `configure.ac` does not require a specific version of glib as of today and glib <2.46 may (or may not) be used on some to-be-supported environments.
  So we either need pre-processor if-else for glib <2.46 and >=2.46 or we need to make glib >=2.46 a known and checked-for build time dependency in `configure.ac`.
  @radosroka any thoughts on these options?
  
- CI does not like the current code formatting, see:
  - log at https://github.com/USBGuard/usbguard/runs/6035863630?check_suite_focus=true#step:7:139
  - helper `./scripts/reformat-sources.sh`

- (My picky vote for `GDBusMessage* const message` adding `const`.)

- (Pull request pull request #545 probably already "fixed" the situation for gnome-settings-daemon (but this change still seems like a good idea).)

What do you think?

Best, Sebastian

muelli:
> * Flag `G_DBUS_MESSAGE_FLAGS_ALLOW_INTERACTIVE_AUTHORIZATION` seems to require glib >=2.46 according to https://libsoup.org/gio/GDBusMessage.html.
>       `configure.ac` does not require a specific version of glib as of today and glib <2.46 may (or may not) be used on some to-be-supported environments.
>       So we either need pre-processor if-else for glib <2.46 and >=2.46 or we need to make glib >=2.46 a known and checked-for build time dependency in `configure.ac`.
>       @radosroka any thoughts on these options?

good catch.
I'd bump the dependency...

hartwork:
> good catch. I'd bump the dependency...

@muelli alright, let's start with that approach then, I guess

muelli:
> * (Pull request pull request #545 probably already "fixed" the situation for gnome-settings-daemon (but this change still seems like a good idea).)

that indeed addresses part of the problem. The other part is that we try to authorise keyboards when the screen is locked. We may or may not be able to do that currently. With this change we can detect whether the user is actually allowed to authorise the device. It would still be better if the user could actually perform the authorisation, but this change probably encodes the right behaviour, because it checks whether the caller is fine with asking for authorisation.

hartwork:
We should should check with say d-feet if interactive auth still works with this patch in practice before merging, to be sure. I haven't done that, yet.

hartwork (changes requested):
@muelli I believe this is the place to insert >=2.46 (for gio): https://github.com/USBGuard/usbguard/blob/4982b018cdda52c539e8b08ae158e54cb3a13e2f/configure.ac#L470

benzea:
Note that while this patch is correct, it causes a timeout error from the client (https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/issues/676#note_1430710). The problem is that in various places there is code like:

```
      if (! isAuthorizedByPolkit(invocation)) {
        return;
      }
```

Which is obviously wrong, as it means that no error is returned. Instead, the code should read

```
      if (! isAuthorizedByPolkit(invocation)) {
        g_dbus_method_invocation_return_error(invocation, G_DBUS_ERROR,
          G_DBUS_ERROR_ACCESS_DENIED, "Could not authenticate request"););
        return;
      }
```

Though, not sure if `G_DBUS_ERROR_ACCESS_DENIED ` or `G_DBUS_ERROR_AUTH_FAILED ` is right. I suppose access denied, unless we pass out information about errors to talk to policykit and then return AUTH_FAILED in that case?

muelli:
> Which is obviously wrong, as it means that no error is returned. Instead, the code needs to call

uh.
thanks for pointing it out.
I know only little about dbus architecture and I am a bit surprised. I expected that kind of error to be produced by the broker, somehow.

But it feels like it would be easy to add.

reg. the type of error: The docs have this to say:
```
G_DBUS_ERROR_ACCESS_DENIED | Security restrictions don't allow doing what you're trying to do.
G_DBUS_ERROR_AUTH_FAILED | Authentication didn't work.
```
(from https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/gio/gio-GDBusError.html - why does it show up under the gstreamer namespace? Rather than a more generic "dbus" namespace or something.)

hartwork:
@benzea thanks for your input, that's definitely helpful! :+1: 

@muelli how should we proceed with this pull request? Should I (a) cherry-pick your commit and make a new pull request applying the ideas of @benzea or (b) make commits for those, export using `git format-patch` and attach those files here for you to add here using `git am` or (c) … — what do you think? I should note that I have no more permissions than you in USBGuard.

benzea:
> I know only little about dbus architecture and I am a bit surprised. I expected that kind of error to be produced by the broker, somehow.

The broker isn't doing the access control, so it can't react anymore.

hartwork:
> > I know only little about dbus architecture and I am a bit surprised. I expected that kind of error to be produced by the broker, somehow.
> 
> The broker isn't doing the access control, so it can't react anymore.

Who is the broker in this picture?

benzea:
> Who is the broker in this picture?

The server that forwards messages between clients. i.e. dbus-daemon or dbus-broker in almost all cases.

hartwork:
Regarding the missing reply aspect brought up by @benzea, there is pull request #548 ready fro review and testing now.

Regarding the allow-interactive-flag aspect brought up by @muelli:
Maybe it's okay that the `dbus-send` CLI never sends the flag with its calls (so that making calls with e.g. `sudo` is required) but GNOME D-Bus GUI tool [d-feet](https://wiki.gnome.org/Apps/DFeet) also doesn't send the flag which means that requiring the flag in USBGuard would break use from d-feet, as of today. I have opened a ticket for clarification with d-feet upstream at https://gitlab.gnome.org/GNOME/d-feet/-/issues/26 now.

I'm looking forward to your feedback :beers:

muelli:
> GUI tool [d-feet](https://wiki.gnome.org/Apps/DFeet) also doesn't send the flag which means that requiring the flag in USBGuard would break use from d-feet, as of today.

That's a problem of the tool, i.e. d-feet.
The correct behaviour is to check whether interactive authorisation was allowed by the caller. If the caller doesn't want popups to be spawned, then it'd be nice if usbguard respected that request.
Whether d-feet is capable of creating such a request is not relevant for encoding the correct behaviour.

Cropi:
Hello @muelli , may I ask you to bump the dependency for gio-2.0 >=2.26 as @hartwork suggested?

hartwork:
@Cropi @muelli >=2.**4**6 please, see https://libsoup.org/gio/GDBusMessage.html#GDBusMessageFlags

muelli:
> Hello @muelli , may I ask you to bump the dependency for gio-2.0 >=2.26 as @hartwork suggested?

I think I tried but couldn't add this new dependency to configure.ac since my autotools-fu is not very good. But I can try again.

hartwork:
@muelli I believe we need something like this:

```diff
   #
   # Check for required D-Bus modules
   #
-  PKG_CHECK_MODULES([dbus], [dbus-1 gio-2.0 polkit-gobject-1],
+  PKG_CHECK_MODULES([dbus], [dbus-1 gio-2.0 >= 2.46 polkit-gobject-1],
   [AC_DEFINE([HAVE_DBUS], [1], [Required GDBus API available])
   dbus_summary="system-wide; $dbus_CFLAGS $dbus_LIBS"],
   [AC_MSG_FAILURE([Required D-Bus modules (dbus-1, gio-2.0, polkit-gobject-1) not found!])]
```

You can try with a version that your system has no way of satisfying — e.g. `100.0` — to see if things fail as expected. Script `./autogen.sh` can be used to regenerated `configure` from `configure.ac` sources.

muelli:
I've added the check now.
It bails out when I put a fantasy value for the version, so I believe it's working as intended.

I haven't tested these changes, though.

muelli:
okay, I've managed to test the change.
The flag is correctly received by USBGuard.
For some reason I don't yet understand, it allows me everything all the time. So there might be glitch in the code, but given that it looks so innocent, I guess it's my system's configuration.

Cropi (approved):

PR #551, opened 2022-08-19
Unreference PolkitAuthorizationResult and PolkitAuthority structs if needed

Cropi:
1. If ```polkit_authority_get_sync()``` fails, then it will return null, so there is no need to free it with ```g_object_unref()```.
2. The same goes for ```polkit_authority_check_authorization_sync()```. The documentation says the following: "A ```PolkitAuthorizationResult``` is returned or NULL if error is set. Free with g_object_unref()."

I came across this issue when SElinux did not allow usbguard-dbus to perform the mentioned actions. Thus, usbguard tried to free memory, which was not even allocated.

[1] https://www.freedesktop.org/software/polkit/docs/latest/PolkitAuthority.html

### DIFF
--- a/src/DBus/DBusBridge.cpp
+++ b/src/DBus/DBusBridge.cpp
@@ -492,12 +492,11 @@ namespace usbguard
     USBGUARD_LOG(Trace) << "Connecting with Polkit authority...";
     PolkitAuthority* const authority = polkit_authority_get_sync(/*cancellable=*/ NULL, &error);
 
-    if (! authority || error) {
+    if (! authority) {
       USBGUARD_LOG(Trace) << "Failed to connect to Polkit authority: " << formatGError(error) << ".";
       *authErrorCode = G_DBUS_ERROR_AUTH_FAILED;
       *authErrorMessage = "Failed to connect to Polkit authority";
       g_error_free(error);
-      g_object_unref(authority);
       g_object_unref(subject);
       return false;
     }
@@ -528,12 +527,11 @@ namespace usbguard
         /*cancellable=*/ NULL,
         &error);
 
-    if (! result || error) {
+    if (! result) {
       USBGUARD_LOG(Trace) << "Failed to check back with Polkit for authoriation: " << formatGError(error) << ".";
       *authErrorCode = G_DBUS_ERROR_AUTH_FAILED;
       *authErrorMessage = "Failed to check back with Polkit for authoriation.";
       g_error_free(error);
-      g_object_unref(result);
       g_object_unref(details);
       g_object_unref(authority);
       g_object_unref(subject);


hartwork (commented):
Hi @Cropi ,

for review of this pull request I asked myself two questions:

- Can function `polkit_authority_get_sync` return `NULL`?
  I found the answer in file `polkit-v.121/src/polkit/polkitauthority.c`: Yes, it can.

- Can function `g_object_unref` be called with `NULL`, safely?
  After some digging, the answer is: No, it cannot.
  
With those two answers I agree that this pull request is moving the code into the right direction. I would suggest to also drop the second `|| error` from `! authority || error` for both consistency and readability, and then I'd consider it "perfect". What do you think?

Best, Sebastian

hartwork:
@Cropi PS: For the [CI failure with Ubuntu 21.10](https://github.com/USBGuard/usbguard/runs/7914037604?check_suite_focus=true), I believe the main point was to cover the most recent Ubuntu, more recent than GitHub CI's `ubuntu-latest`, so maybe we should migrate to Ubuntu 22.04 there, ideally in a dedicated pull request. I'd be happy with either you or me doing a pull request on that matter, and then we could rebase #551 on top of that after merge. What do you think?

Cropi:
> Hi @Cropi ,
> 
> for review of this pull request I asked myself two questions:
> 
>     * Can function `polkit_authority_get_sync` return `NULL`?
>       I found the answer in file `polkit-v.121/src/polkit/polkitauthority.c`: Yes, it can.
> 
>     * Can function `g_object_unref` be called with `NULL`, safely?
>       After some digging, the answer is: No, it cannot.
> 
> 
> With those two answers I agree that this pull request is moving the code into the right direction and fixes a real problem +1 . I would suggest to also drop the second `|| error` from `! authority || error` for both consistency and readability, and then I'd consider it "perfect". What do you think?
> 
> Best, Sebastian

Thanks for the review. I definitely agree that we should change the ```authority``` part as well to ensure better consistency.

hartwork:
@Cropi with PRs #548, #551, #552 all merged now — how do you feel about cutting a release 1.1.2?

Cropi:
> @Cropi with PRs #548, #551, #552 all merged now — how do you feel about cutting a release 1.1.2 off `master`?

I am not against it but usually @radosroka is the one who creates the releases.

hartwork:
@radosroka any chance?

radosroka:
> @radosroka any chance or objections?

I can create release later today.

hartwork:
@radosroka cool! Thank you!

radosroka:
> @radosroka cool! Thank you!

Release is out there:

https://github.com/USBGuard/usbguard/releases/tag/usbguard-1.1.2

hartwork:
@radosroka thank you! [Bumped to 1.1.2 in Gentoo](https://github.com/gentoo/gentoo/commit/4bdb8e4fd557eef61d98fe9155f749309dd250b7) as well now.

PR #555, opened 2022-09-12
scripts: Use bash-builtin `command -v` instead of external `which`

a1346054:
`command -v` is a bash builtin and is a standardized version of `which`

### DIFF
--- a/scripts/copr-build-srpm.sh
+++ b/scripts/copr-build-srpm.sh
@@ -17,8 +17,8 @@
 #
 # Authors: Daniel Kopecek <dkopecek@redhat.com>
 #
-CURL=$(which curl)
-JQ=$(which jq)
+CURL=$(command -v curl)
+JQ=$(command -v jq)
 
 if [ -z "$CURL" ]; then
   echo "Required curl command missing. Aborting."
--- a/scripts/usb-descriptor-collect.sh
+++ b/scripts/usb-descriptor-collect.sh
@@ -19,15 +19,15 @@
 #
 set -e -o pipefail
 
-SED=$(which sed)
-FIND=$(which find)
-LSUSB=$(which lsusb)
-SHA1SUM=$(which sha1sum)
-TAR=$(which tar)
-MKTEMP=$(which mktemp)
-CP=$(which cp)
-WC=$(which wc)
-RM=$(which rm)
+SED=$(command -v sed)
+FIND=$(command -v find)
+LSUSB=$(command -v lsusb)
+SHA1SUM=$(command -v sha1sum)
+TAR=$(command -v tar)
+MKTEMP=$(command -v mktemp)
+CP=$(command -v cp)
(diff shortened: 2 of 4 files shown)


hartwork (changes requested):
Hi @a1346054,

I checked all four files for Bash shebang lines and they all have one (while not the most portable version) so that requirement is filled.  However I'm surprised to see `command -v` in the new version rather than `type -P`, the reason being that if our intention is replacing `which` by something that is equivalent but faster, the fact that `command -v` would also recognize (and prefer) Bash functions of the requested name seems like a mis-feature.  Let me demonstrate:

```console
# command -v curl
/usr/bin/curl

# type -P curl
/usr/bin/curl

# curl() { :; }

# command -v curl
curl

# type -P curl
/usr/bin/curl
```

So I'd consider use of `type -P` more robust a closer match.  What do you think?

Best, Sebastian

PS: This is just my 2 cents and I don't have write/merge permissions in this repository.

hartwork:
> The behavior of `command -v` regarding functions and aliases should not be a problem because user-defined functions and aliases don't persist into scripts.

True, but someone could add functions right into these scripts. I'm not saying it's super likely to happen but `command` is not a fit in that regard. It's not a drop-in replacment to `which`.

> Additionally, `command -v` is not just a bash builtin, it's defined by POSIX, so it behaves the same everywhere.
> 
> `type -P` is not POSIX

Our target is Bash, not POSIX.

a1346054:
> True, but someone could add functions right into these scripts.

then someone could also do `type() { echo "something"; };` and break stuff, so we'd need to protect it by calling:

```sh
builtin type -P curl
```

but it gets really crazy if we assume the scenario that users will edit the provided scripts in crazy ways.

hartwork:
> > True, but someone could add functions right into these scripts.
> 
> then someone could also do `type() { echo "something"; };` and break stuff, so we'd need to protect it by calling:
> 
> ```shell
> builtin type -P curl
> ```

It doesn't seem to stop there, even `builtin() { :; }` seems possible, just tried with Bash 5.1.16.

> but it gets really crazy if we assume the scenario that users will edit the provided scripts in crazy ways.

I was not speaking of end users. I meant people adjusting these scripts in Git upstream over time. 

I think we have both made our points. I would use `type -P` over `command -v` because it's more to the point and because we have Bash features available. We can leave the decision to someone with actual merge permissions. @Cropi @radosroka what do you think?

a1346054:
> I was not speaking of end users. I meant people adjusting these scripts in Git upstream over time.

In that case, it might be of interest to note that the behavior of `command -v curl` is correct there.

If some maintainer for example does:

```sh
curl() { wget "$@"; };
```
with the intention that calling `curl` is now overridden, then `command -v curl` will respect that whereas `type -P curl` will do something else that the maintainer might not have realized.

PR #562, opened 2022-11-01
Specify what happens when neither RuleFile nor RuleFolder is set

Cropi:
Precisely specify what happens when neither RuleFile nor RuleFolder is set. Whilst this is an extraordinary use case, some might be confused in case they encounter the issue.

Reproduction of the issue is straightforward: remove RuleFile and RuleFolder from the config.

### DIFF
--- a/src/Daemon/RuleSetFactory.cpp
+++ b/src/Daemon/RuleSetFactory.cpp
@@ -76,7 +76,7 @@ namespace usbguard
       }
 
       if (ruleSet.empty()) {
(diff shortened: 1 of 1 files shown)


hartwork (approved):
Seems like a good idea :+1: 

Validated based on…

https://github.com/USBGuard/usbguard/blob/927a41da7538d9f31846bdf14a49197d8e5a0d80/src/Daemon/RuleSetFactory.cpp#L66-L76

…and…

https://github.com/USBGuard/usbguard/blob/927a41da7538d9f31846bdf14a49197d8e5a0d80/src/Daemon/Daemon.cpp#L198-L225

PR #572, opened 2022-12-06
Describe comments in the manual page, fixes #461

### LINKED ISSUE #461: Feature Request: Enable comments in rules.conf
A way to add comments to the rules file syntax would be nice, eg: "any line beginning with a hash sign is treated as a comment".

Cropi:
Hi @hartwork, any suggestions on extending the documentation or do you think this will be enough?

### DIFF
--- a/doc/man/usbguard-rules.conf.5.adoc
+++ b/doc/man/usbguard-rules.conf.5.adoc
@@ -245,6 +245,9 @@ List of conditions:
 Partial rule is a rule without a rule target.
 Partial rules may by used by some commands of *usbguard* CLI tool.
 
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@Cropi I think that's okay for a start :+1:

hartwork:
@Cropi I believe I have a fix for the broken CI at #573. The cause is that (1) `ubuntu-latest` is [in a state of migration](https://github.com/actions/runner-images#ongoing-migrations) where it may resolve to either 20.04 or 22.04 and that (2) ldap-utils have removed depracted argument `-h <host>`.

PR #573, opened 2022-12-06
Fix CI by fixing calls to ldap-utils

hartwork:
Related red CI run: https://github.com/USBGuard/usbguard/actions/runs/3627801259/jobs/6118369344

CC @Cropi


Cropi:
Thanks for pointing this out. So the "-h" options seems to be deprecated.

PR #576, opened 2022-12-12
Add "--version" option to the usbguard CLI

### LINKED ISSUE #570: Add option --version
Running `usbguard --version` at least on Ubuntu 22.04 does not produce any versioning information. Handling these issues and fixes might be easier with known version numbers.

So please add support for new option `--version`

Cropi:
Running ```usbguard --version``` should help us handle fixes and issues more easily in the future. By doing so, it's easier to track down which options have been enabled during build time. Fixes #570

Any suggestions on what else to display are welcomed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -433,7 +433,9 @@ usbguard_SOURCES=\
 	src/CLI/usbguard-add-user.cpp \
 	src/CLI/usbguard-add-user.hpp \
 	src/CLI/usbguard-remove-user.cpp \
-	src/CLI/usbguard-remove-user.hpp
+	src/CLI/usbguard-remove-user.hpp \
+	src/CLI/usbguard-print-version.cpp \
+	src/CLI/usbguard-print-version.hpp
 
 usbguard_CXXFLAGS=\
 	$(PTHREAD_CFLAGS)
--- a/scripts/bash_completion/usbguard
+++ b/scripts/bash_completion/usbguard
@@ -251,7 +251,7 @@ _usbguard() {
     # If there was no positional argument provided yet, complete commands
     if [[ $args -eq 1 ]]; then
         opts="get-parameter set-parameter list-devices allow-device block-device reject-device list-rules append-rule"
-        opts="${opts} remove-rule generate-policy watch read-descriptor add-user remove-user"
+        opts="${opts} remove-rule generate-policy watch read-descriptor add-user remove-user --version"
 
     else
         opts='-h --help'
--- a/src/CLI/usbguard-print-version.cpp
+++ b/src/CLI/usbguard-print-version.cpp
@@ -1,0 +1,83 @@
+//
+// Copyright (C) 2022 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+// Authors: Attila Lakatos <alakatos@redhat.com>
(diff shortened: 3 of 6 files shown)


hartwork (changes requested):
Hi @Cropi,

I had a quick look through the changes and also inspected the output locally:

```
# ./usbguard --version
usbguard 1.1.2 compiled with:
        Platform:                       x86_64-pc-linux-gnu
        Platform (lsb_release -d)       Description:    Gentoo Linux
        Linux audit support:            1
        Libcapng support:               1
        Seccomp support:                1
        Crypto backend library:         libsodium
```

Based on what I saw I would like to feedback the following:

- I like the idea of including some additional information to ease reporting.
  At least we can ask for that output from users running a recent enough version.
- Currently, all of that information included is gathered at _compile_ time.
  For two of those, that could be trouble:
  - "Platform"
  - "Platform (lsb_release -d)"

  I suggest to drop these two because of risk of being out of sync with the running
  machine and because gathering them live from C++ code is probably more code
  and effort than healthy here. Also, theyes can be asked for rather easily.
- The `Description:    ` part from output of `lsb_release -d` looks out of place and unintended.
- The new code roughly follows pattern `if foo_enabled: print("foo: 1")`.
  The problem here is that:
  - Absence is not communicated, there will be no line `foo: 0` for absent features. That's rather unpractical.
  - A constant value of 1 does not seem to offer much value.
  I would like to suggest adding `#else` branches and to go to human `enabled`
  and `disabled` or similar.
- The use of tabs in the output seems well intended but does not feel "organic".
  I suggest use of only two spaces rather than tabs for a more "human" feel.
- Adding two new files for the new functions seems overkill, given that the new code could live right next to sibling function `showTopLevelHelp` in file `src/CLI/usbguard.cpp`.
- Bash completion at `scripts/bash_completion/usbguard` should probably learn about `--version` as well.

What do you think?

Best, Sebastian

Cropi:
Hi Sebastian,

> Hi @Cropi,
> 
> I had a quick look through the changes and also inspected the output locally:
> 
> ```
> # ./usbguard --version
> usbguard 1.1.2 compiled with:
>         Platform:                       x86_64-pc-linux-gnu
>         Platform (lsb_release -d)       Description:    Gentoo Linux
>         Linux audit support:            1
>         Libcapng support:               1
>         Seccomp support:                1
>         Crypto backend library:         libsodium
> ```
> 
> Based on what I saw I would like to feedback the following:
> 
>     * I like the idea of including some additional information to ease reporting.
>       At least we can ask for that output from users running a recent enough version.
> 
>     * Currently, all of that information included is gathered at _compile_ time.
>       For two of those, that could be trouble:
>       
>       * "Platform"
>       * "Platform (lsb_release -d)"
>       
>       I suggest to drop these two because of risk of being out of sync with the running
>       machine and because gathering them live from C++ code is probably more code
>       and effort than healthy here. Also, theyes can be asked for rather easily.
> 
I am fine either way. I saw these in another open source project, [rsyslog](https://github.com/rsyslog/rsyslog) , that's why I thought we could give it a try.

>     * The `Description:    ` part from output of `lsb_release -d` looks out of place and unintended.
> 
>     * The new code roughly follows pattern `if foo_enabled: print("foo: 1")`.
>       The problem here is that:
>       
>       * Absence is not communicated, there will be no line `foo: 0` for absent features. That's rather unpractical.
>       * A constant value of 1 does not seem to offer much value.
>         I would like to suggest adding `#else` branches and to go to human `enabled`
>         and `disabled` or similar.
I agree.

> 
>     * The use of tabs in the output seems well intended but does not feel "organic".
>       I suggest use of only two spaces rather than tabs for a more "human" feel.
I agree.

> 
>     * Adding two new files for the new functions seems overkill, given that the new code could live right next to sibling function `showTopLevelHelp` in file `src/CLI/usbguard.cpp`.
> 
>     * Bash completion at `scripts/bash_completion/usbguard` should probably learn about `--version` as well.
Seems good to me.
> 
> 
> What do you think?
> 
> Best, Sebastian

I wonder if there is anything else we could include in the output. Maybe if you build usbguard from sources, then the last commit ID could be mentioned as well.

hartwork:
> I wonder if there is anything else we could include in the output. Maybe if you build usbguard from sources, then the last commit ID could be mentioned as well.

@Cropi sounds good! It's a classic way to break the build for release tarballs without a `.git` folder around but if done properly... :smiley:

PR #582, opened 2023-01-24
Fix compile with GCC 13

yselkowitz:
This is needed for https://gcc.gnu.org/gcc-13/porting_to.html

### DIFF
--- a/src/Library/Base64.hpp
+++ b/src/Library/Base64.hpp
@@ -24,6 +24,7 @@
 
 #include <string>
 #include <cstddef>
+#include <cstdint>
 
 namespace usbguard
 {


hartwork:
@rfrohl the addition is proposed for `Base64.hpp` which is included from `Base64.cpp`. So I believe the PR already does exactly what you want. Could you re-check and help me understand what I may be missing?

rfrohl:
I just wanted to highlight that the include should be moved instead of just adding it to in `Base64.hpp`. 

I only later saw #583, which is the exact change that I think would be the cleanest and it includes the removal of the include in `Base64.cpp`. But both of them fix the issue.

hartwork:
@rfrohl now I understand — thank for the clarification! :+1: 

@Cropi can we merge #583 and close #582 as superseded? I see use of `uint8_t` in `Base64.hpp` so even without GCC 13 around for testing, the pull request seems justified.

PS: @rfrohl what distro are you on, which way do you access GCC 13? I re-check myself just now and I see no GCC 13 at https://hub.docker.com/_/gcc and Gentoo also has GCC 13 hard-masked, still. Just curious.

rfrohl:
> PS: @rfrohl what distro are you on, which way do you access GCC 13? I re-check myself just now and I see no GCC 13 at https://hub.docker.com/_/gcc and Gentoo also has GCC 13 hard-masked, still. Just curious.

openSUSE Tumbleweed switched to gcc13 on the 21th of March [0] and we are cleaning up the last smaller problems. I just realized today that usbguard stopped building and was working on a quick resolution (and therefor was looking if someone already reported/fixed the issue upstream to avoid the issue for other maintainers).

[0] https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/HROWLRAA5ZYK2S6IJ4GMSMPFHCSCNFKG/

Cropi:
Thank you for the PR. I am closing it as it has been fixed in https://github.com/USBGuard/usbguard/pull/586

PR #583, opened 2023-01-27
include missing <cstdint>

kraj:
gcc 13 moved some includes around and as a result <cstdint> is no longer transitively included [1]. Explicitly include it for uint8_t.

[1] https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/Library/Base64.cpp
+++ b/src/Library/Base64.cpp
@@ -22,7 +22,6 @@
 
 #include "Base64.hpp"
 #include <stdexcept>
-#include <cstdint>
 
 namespace usbguard
 {
--- a/src/Library/Base64.hpp
+++ b/src/Library/Base64.hpp
@@ -23,6 +23,7 @@
 #endif
 
 #include <string>
+#include <cstdint>
 #include <cstddef>
 
 namespace usbguard


hartwork:
Hello @kraj we already have #582 on that very front but your version seems to be more complete. @Cropi what do you think?

hartwork (approved):

Cropi:
LGTM! Thanks for the PR.

I just don't understand why github actions have not been executed. Maybe I am missing something.

hartwork:
> LGTM! Thanks for the PR.
> 
> I just don't understand why github actions have not been executed. Maybe I am missing something.

@Cropi not sure, I think that may be a GitHub hick-up. They all ran over in my fork at https://github.com/hartwork/usbguard/actions?query=branch%3Agcc-13-ci .

hartwork:
@Cropi nevermind, mixed up the two PRs, sorry. My CI covers this as a side effect.

Cropi:
Thank you for the PR. I am closing it as it has been fixed in https://github.com/USBGuard/usbguard/pull/586

PR #586, opened 2023-03-28
Fix build for GCC 13 + make GitHub Actions cover build with GCC 13

hartwork:
Sits on top of the original commit from #583 by @kraj, happy to have that merged first.
Alternative to #582.

CC @Cropi @radosroka @kraj @yselkowitz @rfrohl


kraj (approved):
lgtm

rfrohl:
lgtm as well

Issue #475, opened 2021-04-22
Q: Are device ids unique?


muelli:
We're wondering whether we can safely authorise a device by its ID without risking authorising the wrong device.
Let me explain.
We're listening to USBGuard's signals and then determine whether to allow a device based on, e.g. whether the screen is locked. We also want to lock the screen and only authorise a device when the screen has been locked. So we postpone the authorisation of a device until the screen is fully locked.
Now there's a race: the device can disappear until we want to authorise it. That's not a problem in itself, but what we're wondering now is whether the device IDs are guaranteed to be unique or whether they, say, begin with "1" again after having reached `UINT_MAX-2` which would give the attacker control over the device ID.
(yeah, I'm aware that `UINT_MAX` is big and that an attacker would need to increment quite many times. Probably too many times for any attack to be successful, i.e. with 50k re-plugs per second, it would still take a full day).

So what we're looking for, I guess, is some form of guarantee in the API that says something like "yeah, if you want to authorise device `2`, you will only ever authorise this device or nothing". Bonus points if that's also part of the documentation.

This seems to be heavily related to https://github.com/USBGuard/usbguard/issues/39 which would probably avoid the problem altogether.

As a side quest: how does Linux behave if you re-plug a device 1000 times? AFAICS, the device number is a triple digit, i.e. `/dev/bus/usb/001/001`, so does it roll over after 999?

muelli:
> As a side quest: how does Linux behave if you re-plug a device 1000 times? AFAICS, the device number is a triple digit, i.e. `/dev/bus/usb/001/001`, so does it roll over after 999?

I've experimented with qemu: It rolls over after 127.

muelli:
I've done some experiments.
I've modified `RuleSet::assignID()` to return `Rule::LastID - 10 + _id_next++`.

https://github.com/USBGuard/usbguard/blob/ad7a9c60b3185b6761b2a0913f00a80eb41cfc59/src/Library/public/usbguard/RuleSet.cpp#L224-L227

Then I re-plugged a device a few times and observed the daemon's debug output as well as `usbguard list-devices`.

> So what we're looking for, I guess, is some form of guarantee in the API that says something like "yeah, if you want to authorise device `2`, you will only ever authorise this device or nothing". Bonus points if that's also part of the documentation.

The device IDs do indeed roll over.

[...]

PR #600, opened 2023-09-21
ruleset: detect integer overflow of the ID and bail out

muelli:
The check is semantically correct, because some IDs are reserved. Here, we exploit the fact that the LastID is defined to be max_int-2, so when we increment _id_next often enough, we will eventually reach the reserved LastID. If that ID is reached, we bail out to prevent further damage. We could have used the `__builtin_add_overflow` GCC extension but I don't know how compatible it is. Besides, we probably want to prevent the ID spilling into the reserved LastID anyway.

The daemon logs the exception and keeps running. But it appears dysfunctional, i.e. does not notice new device let alone authorise any. That may or may not be the desired behaviour. But it serves as base for discussion and is probably more desired than the behaviour with the overflow.

To test the change, I have the following line in the assignID function:

```
    const auto base = 1 ? (Rule::LastID - 10) : 0 ;
    const auto next_id = _id_next + 1  +  base;
    USBGUARD_LOG(Debug) << "Base: " << 0 << ", Current RuleID: " << _id_next << ", next " << next_id << ", (" << Rule::LastID << ")";
    ...
    _id_next = next_id - base;

```

### DIFF
--- a/src/Library/public/usbguard/RuleSet.cpp
+++ b/src/Library/public/usbguard/RuleSet.cpp
@@ -223,7 +223,13 @@ namespace usbguard
 
   uint32_t RuleSet::assignID()
   {
-    return _id_next++;
+    const auto next_id = _id_next + 1;
+    if (next_id >= Rule::LastID) [[unlikely]]
+        {
+            throw std::out_of_range("Rule ID too high");
+        }
+    _id_next = next_id;
+    return next_id;
   }
 
   void RuleSet::setWritable()


muelli:
ref: https://github.com/USBGuard/usbguard/issues/475

muelli:
@Cropi do you have any opinion on the integer overflow?

Cropi:
Hi @muelli,
Thanks for the PR! I think it would be better to check if the new device ID is already assigned to an existing device. With this approach, we could search for a new ID when a device is inserted. In case there is no free ID (someone is messing with us) then we could throw the exepction as it is now. What do you think?

muelli:
ah, sounds like a valid approach.
More sane, anyway, than a device obtaining the reserved LastID.
I wouldn't know how to test for the availability of an ID.
I also wonder how a device could get an occupied ID, given that the ID is incremented after an assignment.

radosroka:
I think this can be merged for now. @muelli can you rebase PR branch?

muelli:
I've rebased the patch.

hartwork:
@muelli @radosroka this commit and pull request broke the CI on multiple levels:
- it started using unsupported post-C++17 features,
- it changed `return x++` semantics to `return ++x`,
- which in turn broke the test suite (which speaks for the tests), and
- it was not suitably formatted.

As such and with far-from-green CI, **this pull request should not have been merged** in the first place.

In the future, please let's keep CI green, fix red CI [before anything else](https://blog.hartwork.org/posts/my-approach-to-code-review/#B3), and not merge things that are not green. Thank you!

New pull request #628 provides fixes to all of the regression aspects, for a fully green CI. I'm happy to adjust details as needed, I believe it's the right way forward. Please let me know how you feel about it :pray: 

CC @Cropi

PR #621, opened 2024-04-29
Enable RuleFolder by default

PureTryOut:
I don't really know why this was not enabled by default in the first place but this is really useful, also to extend it with distro-specific configuration without touching the upstream files themselves


Cropi:
This is simply because if `RuleFolder` is set but the folder does not actually exist, then `usbguard-daemon` fails to start and that could lead to a not desired state (no device is enabled).

Cropi:
> That... Seems like something that should be fixed rather than disabling the config option 🤔

Okay, I will add a note to Changelog to alert distro maintainers about the change.

PR #627, opened 2024-06-04
Bump c++ version to 20

Cropi:
Currently, the CI is failing because it's not able to parse the new `[[ unlikely ]]` attribute. I am not aware of any other new features the project uses but it might use in the future. We can either remove the attribute or switch to c++20.

Considering that c++20 has been out for a while and compilers already support it, I think it's safe to do.
More info: https://en.cppreference.com/w/cpp/20

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -54,7 +54,7 @@ AX_CHECK_COMPILE_FLAG([-Wno-implicit-fallthrough],
 #
 # Final project CXXFLAGS are set after configure checks.
 #
-CXXFLAGS="-std=c++17 $EXTERNAL_CXXFLAGS"
+CXXFLAGS="-std=c++20 $EXTERNAL_CXXFLAGS"
 CFLAGS="-std=c99 $EXTERNAL_CFLAGS"
 CPPFLAGS="-DHAVE_BUILD_CONFIG_H $EXTERNAL_CPPFLAGS"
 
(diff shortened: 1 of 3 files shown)


hartwork:
@Cropi thanks for the notification. C++20 is younger than 5 years and so requiring C++20 for a software that (1) can easily do without it and (2) is crucial to non-bleeding-edge enterprise I would like to clearly vote against requiring C++20. If you want to make that move, it first needs a list of must-support systems and checking their versions of GCC and Clang and checking the C++20 support of those compilers. Questions like "what is the oldest release of RHEL, Debian, SUSE, Ubuntu that we need to support?" It takes work and before that work the risk is too high in my book.

On a side note I find it concerning that commit 3672e9f84d2ed55994a0df739c4d6d43316c2272 was pushed to `master` despite failing CI. My vote for protecting branch `master` from direct pushes for everyone, hence have all changes go through pull request, to revert that troublesome commit, and to make a pull request introducing a variant of that commit that has all-green CI.

What do you think?

hartwork:
@Cropi I have fixes for all issues with CI now in pull request #628. I am looking forward to your review :pray:

Cropi:
> @Cropi thanks for the notification. C++20 is younger than 5 years and so requiring C++20 for a software that (1) can easily do without it and (2) is crucial to non-bleeding-edge enterprise I would like to clearly vote against requiring C++20. If you want to make that move, in my view it first needs a list of must-support systems and checking their versions of GCC and Clang and checking the C++20 support of those compilers. Questions like "what is the oldest release of RHEL, Debian, SUSE, Ubuntu that we need to support?" It takes work and before that work the risk is too high in my book.
> 

Yes, I fully agree here. It might be an overkill to switch to a newer version when there are no big features depending on it. Thanks for the quick reaction.

> On a side note I find it concerning that pull request #600 was merged to `master` despite failing CI. Getting red CI fixed should be everyone's first priority before merging anything else.
> 
> What do you think?

PR #647, opened 2025-02-06
Add usbguard-tmpfles.conf

Cropi:
At boot time, systemd‑tmpfiles processes this file and creates /var/log/usbguard so that the usbguard service can start successfully.

https://issues.redhat.com/browse/OCPBUGS-49675


radosroka (approved):

cgwalters:
BTW I would also recommend switching over your RPM spec file to `%ghost` this entry, so it is *always* created by systemd-tmpfiles.

PR #648, opened 2025-03-06
Modified USBguard append-rule to add support to the beginning of the …

boring-brother:
Modified USBguard append-rule to add support to the beginning of the configuration file (-a 0 command)
this cmd will failed：
# usbguard append-rule -a 0 "allow 000e:1923"
IPC ERROR: request id=1: Policy append: rule: Invalid parent ID

from RuleSet::appendRule func，it show can support ID 0  to Add to the head of a rule
usbguard-append-rule-a-0

### DIFF
--- a/src/Library/public/usbguard/Policy.cpp
+++ b/src/Library/public/usbguard/Policy.cpp
@@ -60,6 +60,17 @@ namespace usbguard
     USBGUARD_LOG(Trace) << "parent_id=" << parent_id;
     auto rule = std::make_shared<Rule>(_rule);
 
+    if (parent_id == 0) {
+      auto ruleset = _rulesets_ptr.front();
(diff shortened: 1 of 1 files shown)


Cropi:
Hello, I haven't had time to look into this but this might not be as straightforward as it looks like. An attempt has been made to introduce insertRule in https://github.com/USBGuard/usbguard/pull/480.

Cropi:
Thanks a lot for your work on this.
However, I don't think we should merge support for inserting at the beginning of the rule set. Instead, I would be happy to review a patch that implements inserting a rule at any specified position in the list, rather than just at index 0. We would not fix the original problem. For example to insert to the end of a specific ruleset within a file from the rules.d folder.

Issue #549, opened 2022-06-08
Fails to compile with Catch2 >=3.0.1 (which does not have header file catch.hpp)


hartwork:
Quoting https://github.com/catchorg/Catch2/releases/tag/v3.0.1:
> Including `catch.hpp` no longer works

https://github.com/USBGuard/usbguard/blob/4982b018cdda52c539e8b08ae158e54cb3a13e2f/configure.ac#L399

Related downstream report at https://bugs.gentoo.org/850589

PR #659, opened 2025-04-18
tests: add support for Catch2 v3

yselkowitz:
Catch2 v3 no longer has a monolithic <catch.hpp> and also requires linking against compiled libraries:

https://github.com/catchorg/Catch2/blob/v3.0.1/docs/migrate-v2-to-v3.md

This allows either v2 or v3 of system-installed Catch2 to be used.

Fixes: #549


Cropi (approved):
Thanks for the PR. I appreciate the fallback mechanism.
Just a note, I will create an option to exclude test case execution, so packagers can omit it if they don't want to depend on catch.

Issue #654, opened 2025-04-10
Rulesets from rules.d folder are loaded but not applied


Lukas1811:
**TLDR:** Rules are loaded from RuleFolder but not applied or listed when doing `usbguard list-rules`

I have usbguard 1.0.0+ds-2 running on modified version of Debian 11.11 and install my config using a deb package that diverts the /etc/usbguard/usbguard-daemon.conf file to the following config (removed all the comments to keep it short):

```
# RuleFile=/etc/usbguard/rules.conf
RuleFolder=/etc/usbguard/rules.d/
ImplicitPolicyTarget=block
PresentDevicePolicy=apply-policy
PresentControllerPolicy=apply-policy
InsertedDevicePolicy=apply-policy
AuthorizedDefault=none
RestoreControllerDeviceState=false
DeviceManagerBackend=uevent
IPCAllowedUsers=root
IPCAllowedGroups=root
IPCAccessControlFiles=/etc/usbguard/IPCAccessControl.d/
DeviceRulesWithPort=false
AuditBackend=LinuxAudit
AuditFilePath=/var/log/usbguard/usbguard-audit.log
HidePII=false
```

My /etc/usbguard/rules.d/ folder contains six rule files (names are different irl, numbers and extensions are correct):
* 01-rulea.conf
* 02-ruleb.conf 
* 10-rulec.conf
* 20-ruled.conf 
* 30-rulee.conf 
* 99-reject-remaining.conf (only contains `reject`, commented that out already, didn't changed anything)

When I run the daemon the debug output contains the following outputs:
```
[1744290595.693] (i) NSHandler Loading...
[1744290595.693] (i) separator -> :
[1744290595.693] (i) keys:
[1744290595.693] (i) --->usbguard
[1744290595.693] (i) NSHandler Loaded
[1744290595.693] (i) Loading configuration from /etc/usbguard/usbguard-daemon.conf
[1744290595.693] (i) File has correct permissions.
[1744290595.693] (D) ConfigFilePrivate.cpp@155/parse: Parsed: RuleFolder=/etc/usbguard/rules.d/
[1744290595.693] (D) ConfigFilePrivate.cpp@155/parse: Parsed: ImplicitPolicyTarget=block
[1744290595.693] (D) ConfigFilePrivate.cpp@155/parse: Parsed: PresentDevicePolicy=apply-policy
[...]

PR #661, opened 2025-04-24
Added CAP_DAC_READ_SEARCH capability to systemd service file

Lukas1811:
With the old CapabilityBoundingSet the systemd service is not capable of loading (actually searching for) all permissions in the /etc/usbguard/rules.d folder. Adding CAP_DAC_READ_SEARCH allows the daemon running in a service to load rules from the rules folder.

See Issue #654

### DIFF
--- a/usbguard.service.in
+++ b/usbguard.service.in
@@ -6,7 +6,7 @@ Documentation=man:usbguard-daemon(8)
 [Service]
 OOMScoreAdjust=-1000
 AmbientCapabilities=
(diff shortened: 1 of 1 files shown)


Cropi:
Hello,
I am not able to reproduce it on Debian/Ubuntu. I've tried earlier versions as well. This should be a pretty basic use case, so I would expect others to report it but no one has mentioned this is not working properly.
If you could provide me a reproducer that would be great.

Issue #666, opened 2025-05-07
FDInputStream may be leaking fds


semenzato:
Hi, we're seeing the usbguard daemon running out of fds, and `/proc/pid/fd` shows that an extra fd is added every time I unplug and replug a keyboard:
```
lr-x------. 1 root root 64 May  7 12:53 36 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
lr-x------. 1 root root 64 May  7 12:53 37 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
lr-x------. 1 root root 64 May  7 12:53 38 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
lr-x------. 1 root root 64 May  7 12:54 39 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
lr-x------. 1 root root 64 May  7 12:54 40 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
lr-x------. 1 root root 64 May  7 12:54 41 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
lr-x------. 1 root root 64 May  7 12:58 42 -> /sys/devices/pci0000:00/0000:00:14.0/usb3/3-4/descriptors
```
I took a quick look at the source, and at first glance it seems like the `FDInputStream` class doesn't have a destructor, which suggests that the `descriptors` fd may be left open when the class instance is goes out of scope in DeviceBase.cpp::DeviceBase.

Thanks!

semenzato:
I think this may be enough.  I verified that it stops the leak.
```
diff --git a/src/Common/FDInputStream.hpp b/src/Common/FDInputStream.hpp
index 8528e84..8dc0485 100644
--- a/src/Common/FDInputStream.hpp
+++ b/src/Common/FDInputStream.hpp
@@ -57,6 +57,7 @@ namespace usbguard
   {
   public:
     FDStreamBuf(int fd) : fd_(fd) { }
+    ~FDStreamBuf() { close(fd_); }
 
     std::streamsize xsgetn(char* s, std::streamsize n)
     {
```

semenzato:
Would you like me to formally send a patch?  BTW we're patching this for ChromeOS at Google.  https://chromium-review.googlesource.com/c/chromiumos/overlays/chromiumos-overlay/+/6521787

[...]

Issue #663, opened 2025-04-28
Any chance for a new release USBGuard 1.1.4 off of `master`?


hartwork:
Hi!

Is there a chance for a new release off `master` e.g. for the Protobuf v30 fixes and support for Catch2 v3?
A new release would help distros to not need to cherry-pick the post-1.1.3 fixes.

Thanks and best, Sebastian

CC @Cropi

hartwork:
@Cropi please please 🙏

Cropi:
Hello, I will try to spare some time to do the official build soon (I also need it :D). Allow me some time to process it, I am a bit sick right now.

semenzato:
Thank you @hartwork, I hadn't seen this.  Get well soon @Cropi!

Cropi:
Sorry for the delay. The new release is now available, https://github.com/USBGuard/usbguard/releases/tag/usbguard-1.1.4

hartwork:
@Cropi thank you! 👍 👍

PR #667, opened 2025-05-13
FDInputStream: add FDStreamBuf destructor to stop fd leak.

semenzato:
As mentioned in DeviceBase.cpp in setConnectType(), _FDInputStream (stdio_filebuf) is responsible for closing the file descriptor returned by sysfs_device.openAttribute()._.  This patch closes the fd.


semenzato:
Thank you for looking into this and please forgive mistakes---it's my first GitHub pull request... ;)

semenzato:
(I updated this commit with `git push --force` to include the `fixes` clause.)

hartwork (approved):
@semenzato I didn't see this because it didn't reference #666 or CC me anywhere but great to have now, thank you! :+1:

I did find keyword `override` to need >=C++11 only which is good. I'm having trouble finding a definite answer whether `override` marks the new destructor `virtual` implicitly also, so far. I'm approving with that question unanswered.

CC @Cropi

semenzato:
> @semenzato I didn't see this because it didn't reference #666 or CC me anywhere but great to have now, thank you! 👍
> 
> I did find keyword `override` to need >=C++11 only which is good. I'm having trouble finding a definite answer whether `override` marks the new destructor `virtual` implicitly also, so far. I'm approving with that question unanswered.
> 
> CC @Cropi

Hello and sorry for the delay!

Unfortunately my C++ sucks.  However this appears to be correct, `override` implies `virtual` and at least here (Google) both the style guide and the linter recommend to NOT use both.  (To be sure, I also checked that `std::streambuf` is an alias for `std::basic_streambuf`, which defines a virtual destructor.)  So we should be OK!  But I am not sure about backwards compatibility issues, if any.

hartwork:
@semenzato thanks for the update! We have no source suggesting to add both currently and it has been discussed now so everything is ensured to be conscious, so I'm good with moving forward with status quo, personally :+1:

semenzato:
I am not sure if there is anything else I can do here to get this done.  Thanks!

hartwork:
@semenzato there is a comment from @Cropi at https://github.com/USBGuard/usbguard/issues/663#issuecomment-2889849545 that he will need a few days to recover first. I'm hoping for review and a merge from him after his return.

Cropi:
I think the use of override is fine as `std::streambuf`... indeed has a virtual destructor. I think it's mainly cosmetics, but still :smiley: 
What if `HAVE_EXT_STDIO_FILEBUF_H` is defined and we have a different implementation of `FDInputStream`, don't we need to update it as well?

semenzato:
I wish I could be of more help, but my C++ knowledge is also quite superficial.  I'll see if I can find the time to dig in a bit more.

I haven't tried to build this under different configurations (such as having HAVE_EXT_STDIO_FILEBUF), so I suppose that this patch could break those.  I wonder if there is some kind of automated pre-submit or post-submit test?  Or would the reviewer/maintainer ask me to do that?

The other question is whether this doesn't break anything now, but it's a kludge which might require changes in future configurations.

Is it common to get no responses from the maintainer for weeks?  What happens when the maintainer stops maintaining?

Thanks!

PR #672, opened 2025-09-01
Fix uutils install incompatibility

ReillyBrogan:
The install binary provided by uutils-coreutils does not permit the mode to be specified twice. `$(INSTALL_DATA)` resolves to `/usr/bin/install -c -m 644` on my system, which causes the `install-tmpfiles` target to run `/usr/bin/install -c -m 644 -m 640` which causes `install` to throw an error. As there is no security benefit to removing the read permission from all for tmpfiles configs just remove the `-m 640` to fix that.

While we're at it also use `$(INSTALL_DATA)` in a couple other places that could use it.


Cropi (approved):
Thanks for the PR!

PR #687, opened 2026-03-11
Library: fix race conditions in device management

reinauer:
The _sysfs_path_to_id_map and _backlog structures were being accessed
concurrently by the main thread and the uevent thread without proper
synchronization. This could lead to memory corruption or erroneous
device rejections during the initial system scanning phase.

Fix this by:
1. Protecting _sysfs_path_to_id_map in DeviceManagerBase using the
   existing refDeviceMapMutex().
2. Protecting the event _backlog in UEventDeviceManager using a
   dedicated static mutex to ensure thread-safe buffering and handoff.

### DIFF
--- a/src/Library/DeviceManagerBase.cpp
+++ b/src/Library/DeviceManagerBase.cpp
@@ -95,6 +95,7 @@ namespace usbguard
 
   uint32_t DeviceManagerBase::getIDFromSysfsPath(const std::string& sysfs_path) const
   {
+    std::lock_guard<std::mutex> lock(const_cast<DeviceManagerBase*>(this)->refDeviceMapMutex());
     uint32_t id = 0;
 
     if (knownSysfsPath(sysfs_path, &id)) {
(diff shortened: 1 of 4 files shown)


Cropi (changes requested):
Good direction, a few issues to address:

**1. Deadlock in `DeviceManagerBase`**
`getIDFromSysfsPath` and `isPresentSysfsPath` lock `refDeviceMapMutex()` then call `knownSysfsPath()`, which locks the same mutex again. `std::mutex` isn't recursive so this deadlocks. `knownSysfsPath` should be an internal unlocked helper.

**2. Exception safety in backlog locking**
Raw `pthread_mutex_lock`/`unlock` around `emplace_back` isn't safe — if it throws, the mutex stays locked. Use a member `std::mutex` with `std::lock_guard` instead.

**3. Static mutex over instance data (minor)**
`G_backlog_mutex` is static but `_backlog` is per-instance. No real issue since there's only ever one instance, but a member mutex would be cleaner.

PR #697, opened 2026-07-23
Re-enable GCC 9.x Debian CI

hartwork:
@Cropi what do you think?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,7 +34,7 @@ jobs:
         linux_distro:
         - Alpine Linux 3.21  # with musl
         - CentOS 8.2         # with GCC 8.5.0
-        # - Debian Buster with GCC 9.2  # stock buster has GCC 8.3 - temporarily disabled
+        - Debian Bullseye with GCC 9.5  # stock bullseye has GCC 10.2
         - Ubuntu 22.04       # because super popular
         - Ubuntu 25.04       # because GCC 15+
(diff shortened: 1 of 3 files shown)


hartwork:
@Cropi any thoughts?

Cropi:
@hartwork  aren't the version check lines missing a closing quote?  `[[ "$(gcc -dumpversion) == 9.5.* ]]` should be `[[ "$(gcc -dumpversion)" == 9.5.* ]]` ? The CI passed though. Also, on GCC 9+ `gcc -dumpversion` might return only the major version (`9`), which won't match the `9.5.*` glob.  Should `gcc -dumpfullversion` be used instead?

hartwork:
@Cropi thanks for looking into it!

> @hartwork aren't the version check lines missing a closing quote? `[[ "$(gcc -dumpversion) == 9.5.* ]]` should be `[[ "$(gcc -dumpversion)" == 9.5.* ]]` ? The CI passed though.

That's an interesting find — thanks! Fixed in a follow-up commit now.

> Also, on GCC 9+ `gcc -dumpversion` might return only the major version (`9`), which won't match the `9.5.*` glob. Should `gcc -dumpfullversion` be used instead?

I checked, it's returning `9.5.0` with this precise Docker image (`gcc:9.5-bullseye`). I believe we're good with plain `-dumpversion`.

Cropi (approved):

PR #700, opened 2026-09-09
Initialize the default policy target

Sovtouch:
The Policy constructor leaves _defaultTarget uninitialized. Calling getDefaultTarget() before setDefaultTarget() reads an indeterminate value.

Initialize _defaultTarget to Rule::Target::Block in the constructor to ensure a defined default policy.


Sovtouch:
ping!

